PR #60, opened 2021-08-24
Update README.md

xsnoopy:
Usually NC are connected from high voltage to the NC switch, then to the Input of a controll unit. For Microcontroller you need to wire them towards ground. This should help people to understand that better.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 grblHAL has [many extensions](https://github.com/grblHAL/core/wiki) that may cause issues with some senders. As a workaround for these a [compile time option](https://github.com/grblHAL/core/wiki/Changes-from-grbl-1.1#workaround) has been added that disables extensions selectively. 
 
 __IMPORTANT!__ grblHAL defaults to normally closed \(NC\) switches for inputs, if none are connected when testing it is likely that the controller will start in alarm mode.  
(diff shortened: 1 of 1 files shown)


terjeio (changes requested):
This is a tad more complicated and depends on the configuration. Both NC and NO switches may be wired to the +ve supply. The pullup resistors can be turned off by $18 (this will enable pulldown instead for most, if not all, drivers) and the inputs inverted by $15.

It is also possible to wire limit switches (with three terminals) to GND, the +ve supply and the inputs.

Relying only on the internal microcontroller weak pullup (or pulldown) resistor only for a board design is, IMO, bad practice. Typically these are around 50 K, with a 3.3V +ve supply the current provided could be below the switch [wetting current](https://en.wikipedia.org/wiki/Wetting_current)? A capacitor and/or a lower value pullup/pulldown resistor should be added to the circuit - both for provinding enough current and improve noise immunity. Even better is to use optocouplers (some boards do) - this will increase the switch current, provide better noise immunity (since most are acting as a low-pass filter) _and_ provide protection for the microprocessor pins.

IMO the sentence should be written a bit differently to take this into account.

xsnoopy:
Understood, I was just focusing on the struggle I had. It is described pretty well with your explanation and the regular grbl wiki. I will see that I get something together from your answer and the grbl wiki fitting for grblHAL.

phansel:
should this PR be closed given it is 5 years obsolete and incomplete?

PR #68, opened 2021-09-05
Commit tool table NVS fix.

andrewmarles:
When I went to compile with the tool table enabled, initially it seemed to break the build.  The changes below seemed to fix things for me and allow me to test the functionality.

This was on the teensy4.1 driver with my GRBLHAL2000 board, but I think this would apply to all drivers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,8 @@ It has been written to complement grblHAL and has features such as proper keyboa
 
 ---
 
-Latest build date is 20210803, see the [changelog](changelog.md) for details.
+Latest build date is 20210819, see the [changelog](changelog.md) for details.  
+__NOTE:__ Drivers built with more than three axes configured \(`N_AXIS` > `3`\) will force a settings reset when upgraded. Backup and restore of settings is recommended for these.
 
 ---
 
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,21 @@
 ## grblHAL changelog
 
+Build 20210819:
+
+Core:
+* Added `$376` setting for designating ABC-axes individually as rotational.  
+__NOTE:__ This setting is only available when N_AXIS is > 3 and will force a settings reset on an upgrade for such configurations. Backup and restore settings when upgrading!  
+Scaling from inches to mm is disabled for axes designated as rotational, no other processing takes place.
+* Added `$ESG` and `$ESH` system commands for outputting current setting definitions in [Grbl csv-format](https://github.com/gnea/grbl/tree/master/doc/csv) and grblHAL tab-format respectively.
+Only settings valid in the active configuration will be outputted, driver and plugin specific settings will be added as well - even from well behaved third party code.
+* Added setting descriptions to most core/driver/plugin settings. Third party drivers and plugins may also add descriptions to any settings implementented.  
+NOTE: Drivers for processors with limited flash may not have the descriptions compiled in.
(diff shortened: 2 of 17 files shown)


terjeio:
How did it break the build?

Depending on the number of tools defined this change will potentially overwrite data in the NVS_ADDR_PARAMETERS block (used for offsets). It will also break the assert in grbllib.c:

https://github.com/grblHAL/core/blob/384fda6e3a135f6db1050f55a24e93260528111f/grbllib.c#L94

I have been thinking of adding the tool table to the end of the first 1K block in order to get more headroom. This can be done increasing the GRBL_NVS_SIZE define.

andrewmarles:
Executable went from 500k to 70k and would not run.  I don’t have debugging on teensy so didn’t see where it was failing to load.  I probably don’t know enough about the memory map to understand the best place to put things, still learning.

terjeio:
What did you set N_TOOLS to?

andrewmarles:
Left it at 8 for now.

terjeio:
`N_TOOLS  8` lead to a failed build with the original code? It works ok for me:
```
Memory Usage on Teensy 4.1:
  FLASH: code:159876, data:39520, headers:8472   free for files:7918596
   RAM1: variables:131776, code:157256, padding:6584   free for local variables:228672
   RAM2: variables:12384  free for malloc/new:511904
```
Without tool table:
```
Memory Usage on Teensy 4.1:
  FLASH: code:158852, data:39336, headers:8656   free for files:7919620
   RAM1: variables:127680, code:156232, padding:7608   free for local variables:232768
   RAM2: variables:12384  free for malloc/new:511904
```

This is with ethernet and a few other plugins enabled. And with the 1.54 version of the Teensy libraries - earlier versions are buggy.

terjeio:
I've added the tool table to NVS debug output - from the `$Q` command. With original code and 8 tools:
```
[MSG:NVS Area: addr size]
[MSG:Global: 1 341]
[MSG:Tool table: 343 168]
[MSG:Parameters: 512 156]
[MSG:Startup block: 799 142]
[MSG:Build info: 942 71]
[MSG:Driver: 1024 141]
```
No room  for expansion in the global area. This means I have to move the tool table above 1K...
Note that this is with 3 axes enabled - with four or more axes 8 tools won't fit.

With your change:
```
[MSG:NVS Area: addr size]
[MSG:Global: 1 341]
[MSG:Tool table: 680 168]
[MSG:Parameters: 512 156]
[MSG:Startup block: 799 142]
[MSG:Build info: 942 71]
[MSG:Driver: 1024 141]
```
It goes clear of the parameters area but overlaps the startup area.

Debug output has to be enabled in config.h to enable `$Q`.

andrewmarles:
I've closed this pull request as it obviously isn't the correct solution, but I'm going to continue looking at this because I still cannot get the tool table to compile into my build, nor is it really feasible to only have it working with 3 axes.

terjeio:
Can you try with this patch? It moves the tool table to 1024.

[MSG:NVS Area: addr size]
[MSG:Global: 1 341]
[MSG:Parameters: 512 156]
[MSG:Startup block: 799 142]
[MSG:Build info: 942 71]
[MSG:Tool table: 1024 168]
[MSG:Driver: 1192 91]

[nvs.zip](https://github.com/grblHAL/core/files/7240651/nvs.zip)

PR #77, opened 2021-10-16
Commit conditional compile flags for feed hold and cycle start

andrewmarles:
For when these signals are not used by a board.


terjeio:
I have made the needed changes - will soon commit. I'll merge the PicoBOB PR when you have confirmed it.

PR #81, opened 2021-11-04
Fix for overrides wrapping from minimum to maximum rate

dresco:
Overrides can wrap around to the maximum rate if multiple CMD_OVERRIDE_xx_MINUS commands are queued while
already at, or near to, the minimum rate. This can occur when using an encoder if moved too quickly.

This happens because the arguments to plan_feed_override() and spindle_set_override() are unsigned. If
(for instance) the override rate is 10, and commands from 11 negative encoder ticks are enqueued before
processing, then the requested override wraps round to a large positive number, and is capped at the
max override value.

### DIFF
--- a/planner.c
+++ b/planner.c
@@ -565,7 +565,7 @@ void plan_cycle_reinitialize ()
 }
 
 // Set feed overrides
-void plan_feed_override (uint_fast8_t feed_override, uint_fast8_t rapid_override)
+void plan_feed_override (int_fast16_t feed_override, int_fast16_t rapid_override)
 {
     if(sys.override.control.feed_rate_disable)
         return;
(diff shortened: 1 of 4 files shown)


terjeio:
Hmm, this should be handled in [protocol.c](https://github.com/grblHAL/core/blob/4a140576a2acf12172ef3532b16b433a07984f71/protocol.c#L515-L633) instead?

dresco:
Okay I see, and perform the min/max bounds check at each iteration here instead?

terjeio:
I would read the current value into an int32_t and call the override functions like this:

`spindle_set_override(last_s_override < 0 ? 0 : (last_s_override > 255 ? 255 : (uint_fast8_t)last_s_override));`

and let spindle_set_override() handle the max/min bounds. Same for the feed rate override. I can do this if you don't mind.

terjeio:
nah, you are right - the values should be capped in each iteration...

dresco:
Okay cool, want me to do this & update the PR? Or happy for you to, whatever suits.. Cheers.

terjeio:
I'll make the change, faster for for both me and you. Thanks for discovering this issue.

PR #89, opened 2021-11-20
Initial support for spindle power data

dresco:
Add spindle_data_t members to support spindle power reporting

### DIFF
--- a/spindle_control.h
+++ b/spindle_control.h
@@ -66,6 +66,9 @@ typedef struct {
     float rpm_high_limit;
     float angular_position; // Number of revolutions since last reset
     float rpm_programmed;
+    float amps;
+    float volts;
+    float kw;
     uint32_t index_count;
     uint32_t pulse_count;
     uint32_t error_count;


terjeio:
Are these new fields really needed in the core? Better to have a new structure in the spindle plugin folder, perhaps in a new file named _spindle.h_ so it can be shared between VDF implementations? Reporting could be added in a _spindle.c_ file in the same folder so it can be shared.

Similar to how it is done the [Trinamic top level code](https://github.com/grblHAL/Plugins_motor/blob/master/trinamic.h)?

dresco:
> Are these new fields really needed in the core?

No possibly not, just that the rpm was already being stored in spindle_data_t.. I will have a look at the Trimanic code for ideas, thanks.

dresco:
Hi Terje, seeing your spindle re-work reminded me of this old PR. Happy to close or amend it as needed, but could do with some guidance first..

My goal with this was to display the actual spindle power in my panel plugin, appreciate it may be useful in a realtime report also.

My current code has some changes to the huanyang driver to retrieve the additional registers. I've now added just a watts member to the HAL spindle_data_t struct, and am accessing it in my panel plugin like this;

```
     if(!hal.spindle.get_data) {
         spindle_rpm = lroundf(sp_state.on ? sys.spindle_rpm : 0);
+        spindle_power = 0;
     } else {
         spindle_rpm = lroundf(hal.spindle.get_data(SpindleData_RPM)->rpm);
+        spindle_power = lroundf(hal.spindle.get_data(SpindleData_RPM)->watts);
     }
```
Am not sure how I could achieve this without a change to the core? Or are you suggesting the core just gets a pointer to a new data structure in the spindle plugin, instead of getting the actual field(s)?

terjeio:
You can add to the real-time report in the plugin by hooking code into the `grbl.on_realtime_report` function pointer, so there is no need to modify anything in the core. [Here](https://github.com/grblHAL/Plugin_SD_card/blob/f0b3f01a8120c1598282e7cd0da845ac5346db14/sdcard.c#L450-L468) is an example of how I have done it in the SD card plugin.

dresco:
Thanks, that makes complete sense for getting it to a sender. Am still struggling to see how I would access vfd plugin data from a different plugin though?

terjeio:
> Am still struggling to see how I would access vfd plugin data from a different plugin though?

Define the data structure in a .h and declare it as extern, then include the .h in your plugin. Or add a function that returns a pointer to the data structure and declare that in the .h.

dresco:
Ok, thanks.

PR #123, opened 2022-03-09
Add definition for STM32H743xx series processors

dresco:
I've started a port for the STM32H7xx series processors, based on your F7 driver. This request just adds the 743xx to the core as a known STM32 platform.

### DIFF
--- a/platform.h
+++ b/platform.h
@@ -21,7 +21,7 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


dresco:
Btw, my repo is [here](https://github.com/dresco/STM32H7xx)

Have only tested basic UART and USB-CDC operation so far - no plugins - but seems to be booting and working fine on the WeAct [MiniSTM32H7xx](https://github.com/WeActTC/MiniSTM32H7xx) board.

I have a Nucleo-H743 on back-order, but the expected shipping date has just changed from next week to next year!!  Was planning on looking at ethernet once that arrives - looks a bit more of a challenge (changes in the generated support code vs the F7 libraries).

PR #181, opened 2022-09-02
Allow enqueuing gcode outside of normal input stream when in alarm state

dresco:
Adds STATE_ALARM to the valid states for allowing gcode from outside of the normal input stream. 
I've been using this to allow my panel plugin to unlock the controller from an alarm state (typically after starting the controller with motor power disabled).

### DIFF
--- a/protocol.c
+++ b/protocol.c
@@ -68,8 +68,8 @@ static void protocol_execute_rt_commands (void);
 bool protocol_enqueue_gcode (char *gcode)
 {
     bool ok = xcommand[0] == '\0' &&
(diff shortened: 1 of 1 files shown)


terjeio:
You can execute system commands such as `$X` by calling [system_execute_line()](https://github.com/grblHAL/core/blob/66f821d8e08996fc290c6d7b2321a2742c54569a/system.c#L303-L304) directly. I would prefer that you do that instead of modifying the enqueue function. If changing the enqueue function anyway it should return false for non-system commands (including jog commands) when in alarm state.

dresco:
> You can execute system commands such as `$X` by calling system_execute_line() directly.

Many thanks, I'll try that instead..

PR #266, opened 2023-03-11
Add init for 3rd party control panel plugin

dresco:


terjeio:
I am wondering a bit how to handle plugins that overlap...
This overlaps with KEYPAD_ENABLE, DISPLAY_ENABLE and possibly ENCODER_ENABLE as well.
How to handle this in the Web Builder without ending up in a mess is my main concern, but also that many _my_machine.h_ files needs to be kept in sync with new plugins.

Thinking aloud, would it be acceptable that your plugin get numbers assigned for KEYPAD_ENABLE and DISPLAY_ENABLE, e.g. 4 for both? This way I can keep the Web Builder UI down to two drop-downs, one for keypad and one for display. These can be synced so that when your panel is selected in one the other follow.

BTW the other symbols required can be added to [driver_opts.h similar to this](https://github.com/grblHAL/core/blob/71aa2a426cdd87f39a91124bef5de94585cc7919/driver_opts.h#L135-L140):

```
#if KEYPAD_ENABLE  == 4 || DISPLAY_ENABLE == 4
#undef KEYPAD_ENABLE // allowing you to have only display_init() implemented in your plugin.
MODBUS_ENABLE=1
MODBUS_MAX_ADU_SIZE=64
#endif
```

What do you think?

dresco:
> What do you think?

I think from a UX perspective, it would be nice to have a dropdown in the web builder that allows a choice between the plugins - however that is implemented..

I agree it does overlap with all three. In theory a panel h/w implementation could provide just one aspect of encoder/keypad/display, or in any combination. I don't see a use case where it would co-exist with the other plugins though, so off the top of my head, if it's chosen it could safely #undef the rest.

I have a prototype board that sits in a bezel behind a 3.5" display. I'm just doing a minor h/w revision to tidy a couple of bits up, but will send you one when ready..

dresco:
Hi Terje, ignoring the init question for a bit (coz I haven't thought about it much more)...

I've just pushed a bunch of updates to my [Plugin_panel](https://github.com/dresco/Plugin_panel) repo. Next up on my list is to replace some hard coded values with NVS settings. 

Looking at settings.h, if I want a new settings group, what's the process? It doesn't look like there are any gaps below the Axis groups (which say they must be sequential & last)?

Thanks, Jon.

terjeio:
>  It doesn't look like there are any gaps below the Axis groups (which say they must be sequential & last)?

Where did I say that? New settings can be added before [Setting_SettingsMax](https://github.com/grblHAL/core/blob/334ca8f8fadecbda0a7243fed2c03d3135cce378/settings.h#L358-L360) - plenty of room there, just continue from the last one (533). If you need a block of "numbered" settings, [like this](https://github.com/grblHAL/core/blob/334ca8f8fadecbda0a7243fed2c03d3135cce378/settings.h#L346-L353), then create a hole by starting from 540.

I have to take a look your CAN API. I am thinking about designing a breakout board for Nucleo-144, 8-axis support and aiming to use all the remaining pins for useful stuff. A CAN interface will be part of that.

dresco:
> Where did I say that? 

Here, was the settings group that I was wondering about..
https://github.com/grblHAL/core/blob/334ca8f8fadecbda0a7243fed2c03d3135cce378/settings.h#L712-L735

I've just finished a new revision of my control panel board & associated panel h/w. Just cleaning the code up (starting with the plugin), before seeing if there is any interest.. I'd like to send you a built up front panel to play with if that's okay?

PXL_20230524_164424455

> I have to take a look your CAN API. I am thinking about designing a breakout board for Nucleo-144, 8-axis support and aiming to use all the remaining pins for useful stuff. A CAN interface will be part of that.

Cool, I've finally managed to get a couple of H7 chips, so planning on making a H7 controller board prototype next..

terjeio:
I am mixing stuff up - was thinking settings numbers... You can add groups after `Group_Embroidery`, but update the comments for all the following accordingly. The comments are for the doxygen documentation.

>  I'd like to send you a built up front panel to play with if that's okay?

That would be generous of you so what else can I say but yes and thank you for the offer!
It looks professionally made, anodized and silk screened front panel?
I guess it should not be too hard to get it working with some of the other controllers, at least those available via the Web Builder.

> I've finally managed to get a couple of H7 chips, so planning on making a H7 controller board prototype next..

Nice, do you plan to add a ton of features as well?

dresco:
> You can add groups after Group_Embroidery, but update the comments for all the following accordingly. 

Will do, thanks.

> It looks professionally made, anodized and silk screened front panel?

Thanks, just FR-4 PCB. Cost effective, and quite pleased with how it came out..

> Nice, do you plan to add a ton of features as well?

Just a mental wish list at the moment, but will try and make the most of the features/pins;

image

PR #368, opened 2023-09-16
Allow driver to be considered ok even if only update_rpm is added

infnorm:
Not sure if this is correct, but the simulator for example only initializes update_rpm with certain flag settings.
There seem to be places where update_pwm and update_rpm are checked independently @terjeio?


terjeio:
A better test would be `spindle->get_pwm ? (spindle->update_pwm != NULL) : (spindle->update_rpm != NULL)`?
For now I have solved the issue by instantiating the null spindle in the simulator driver, I'll revisit this a bit later as I will be looking into adding support for stepper driven spindles and completing multiple simultaneously active spindles handling.

PR #479, opened 2024-03-29
disable unneeded settings

heeplr:
* disable ```Setting_SpindleOnDelay``` when no spindle is configured
* disable ```Setting_HoldCoolantOnDelay``` when there are no coolant pins

I disabled a lot of stuff to gain AUX pins and noticed these two settings with $ESH that are unused.

### DIFF
--- a/settings.c
+++ b/settings.c
@@ -2012,9 +2012,11 @@ static bool is_setting_available (const setting_detail_t *setting)
             available = spindle_get_caps(true).at_speed || hal.driver_cap.spindle_encoder;
             break;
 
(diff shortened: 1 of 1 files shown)


terjeio:
Sorry, this cannot work - and changes to the HAL _all_ drivers are required...
The reason is that the core normally does not have access to driver level #defines and has to get the required information from HAL flags, [such as these](https://github.com/grblHAL/core/blob/d9ed1bd89cdd7ea5dff3a1b217912d2447edb0cc/hal.h#L46-L74), or function pointers beeing NULL or not.

So I will make the required changes for you, likely in the next commit.

PR #565, opened 2024-08-04
Bugfix mc_arc multi turns clockwise in motion_control.c

Holoratte:
Bugfix mc_arc multi turns clockwise

since angular_travel is calculated only ccw, sign needs to be changed for the calculation of linear_per_turn.
Otherwise clockwise plane.axis_linear overshoots by most 1 turn, then returns.

### DIFF
--- a/motion_control.c
+++ b/motion_control.c
@@ -247,9 +247,7 @@ void mc_arc (float *target, plan_line_data_t *pl_data, float *position, float *o
     if(labs(turns) > 1) {
 
         uint32_t n_turns = labs(turns) - 1;
-        float arc_travel = 2.0f * M_PI * n_turns - angular_travel;
-		if (turns > 0)
-			arc_travel = 2.0f * M_PI * n_turns + angular_travel;
+        float arc_travel = 2.0f * M_PI * (float)n_turns + (turns > 0 ? angular_travel : -angular_travel);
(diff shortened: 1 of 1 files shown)


motion_control.c
```diff
@@ -247,7 +247,9 @@ void mc_arc (float *target, plan_line_data_t *pl_data, float *position, float *o
     if(labs(turns) > 1) {
 
         uint32_t n_turns = labs(turns) - 1;
-        float arc_travel = 2.0f * M_PI * n_turns + angular_travel;
+        float arc_travel = 2.0f * M_PI * n_turns - angular_travel;
+		if (turns > 0)
+			arc_travel = 2.0f * M_PI * n_turns + angular_travel;
```
terjeio, line 252: IMO better to use the ternary operator, likely results in less code emitted (the STM32F1xx driver is tight on flash):

` float arc_travel = 2.0f * M_PI * (float)n_turns + (turns > 0 ? angular_travel : -angular_travel);`

terjeio: > At least for Teensy 4.1 using Arduino IDE2 that does not matter

Ok, then the compiler optimizes away the duplicated parts of the code - not bad. Anyway, IMO by using ternary operator it makes the code simpler and a bit easier to read.

PR #589, opened 2024-09-26
Fix memory leak

skasti:
Hoping this fixes the memory leak I caused. Sorry about that 😅


skasti:
Feel free to dismiss this PR if you have already started fixing this 😅

PR #593, opened 2024-09-29
Jerk Acceleration Settings and HAAS G187 Acceleration Profiles

Dietz0r:
This adds code to implement 3rd order accleration settings over the acceleration ticks of the stepper loop.

With the default settings of 100 acceleration ticks it will update the acceleration values according to the jerk setting with each 10ms block that gets fed into the stepper loop.
If a CPU is strong enough to run 1000 acceleration ticks that would net a 1ms Jerk Stepper Loop but 100 or 200 ticks work fine.
Tested on 2 independent machines with help from @empyreancnc

Known Issue: 
The machine moves a bit erratically/stuttery when getting fed jog commands from either a pendant or the IOSender interface, i think this has to do with how those conitinous jog commands are fed to the controller but i wasnt able to veryify or fix the issue on my end. Normal G-Code file execution works as expected. I hope you might have more insight into why this problem arises.

Also added support for HAAS style G187 acceleration profiles currently set to 20%, 40% 60% 80% and 100% of the max settings to allow for fine control during toolpathes and improved surface finishes. 
Gets called as G187 P1 for 100% speed roughing to G187 P5 for 20% slow finishing.

Hope this helps! :)


Dietz0r:
No questions, annotations, comments or change requests? =(

terjeio:
I'll get to this when I am back home early next month since I do not have access to a machine where I am now.

terjeio:
FYI I'll be back home in a week...

From the [Haas documentation](https://www.haascnc.com/service/codes-settings.type=gcode.machine=mill.value=G187.html):

> G187 will be canceled whenever [RESET] is pressed, M30 or M02 is executed, the end of program is reached, or [EMERGENCY STOP] is pressed.

This means that there should be a setting for the default P value? And cancellation implemented?

gcode.c
```diff
@@ -2436,7 +2445,18 @@ status_code_t gc_execute_block (char *block)
                     gc_block.values.xyz[idx] = gc_state.g92_coord_offset[idx];
             } while(idx);
             break;
-
+            
+#if ENABLE_ACCELERATION_PROFILES
+        case NonModal_SetAccelerationProfile:
+            if (!gc_block.values.p){
```
terjeio, line 2451: This should be !gc_block.words.p to check if a value has been supplied.
IMO also check gc_block.words.e, if supplied (true) error out since it is not supported?

settings.h
```diff
@@ -1064,6 +1071,11 @@ bool settings_read_coord_data(coord_system_id_t id, float (*coord_data)[N_AXIS])
 // Temporarily override acceleration, if 0 restore to configured setting value
 bool settings_override_acceleration (uint8_t axis, float acceleration);
 
+#if ENABLE_ACCELERATION_PROFILES
+extern uint8_t ActiveAccelProfile;
```
terjeio, line 1075: IMO ActiveAccelProfile should be added to gc_state and passed to the planner via pl_block instead. Set the default in gc_init() and restore it in the M2/M30 handling code.

Dietz0r:
> FYI I'll be back home in a week...

Hope you enjoyed your trip! :) 

> This means that there should be a setting for the default P value? And cancellation implemented?

I think the most reliable thing would be to have the default value be 100%, so the settings that are set on the axis and be modified from there with the profiles. Saves a setting slot and if you really want to just run at 80% max it's trivial to just add G187 P4 at the start of any code.

The cancellation is a good point! didnt think of that! Will look at it and add it.

> This should be !gc_block.words.p to check if a value has been supplied. IMO also check gc_block.words.e, if supplied (true) error out since it is not supported?

So no one accidentally tries to use a real HAAS routine on it? yeah i see it.

settings.c
```diff
@@ -772,6 +802,13 @@ PROGMEM static const setting_descr_t setting_descr[] = {
     { (setting_id_t)(Setting_AxisStepsPerMM + 1), "Travel resolution in steps per degree." }, // "Hack" to get correct description for rotary axes
     { Setting_AxisMaxRate, "Maximum rate. Used as G0 rapid rate." },
     { Setting_AxisAcceleration, "Acceleration. Used for motion planning to not exceed motor torque and lose steps." },
+#if ENABLE_JERK_ACCELERATION   
+    { Setting_AxisJerk, "Maximum rate of acceleration change - smoothes out acceleration profile up to max axis acceleration.\\n\\n"
+                        "Minimum value of x10 Acceleration setting to ensure decent acceleration times.\\n"
+                        "Maximum is calcualted by current acceleration and stepper segment time.\\n"
```
terjeio, line 808: calcualted - > typo

settings.c
```diff
@@ -889,12 +926,28 @@ bool settings_override_acceleration (uint8_t axis, float acceleration)
...
     }
-
     return true;
 }
 
+#if ENABLE_ACCELERATION_PROFILES
+ActiveAccelProfile = 1; // Initialize machine with 100% Profile
+
+//Acceleration Profiles for G187 P[x] in percent of maximum machine acceleration.
+float LookupProfile(uint8_t Profile) {
```
terjeio, line 938: Move to planner.c as static inlined?
and the return value is either 0 or 1 since you assign the value to Profile before returning it? Use return lookup(profile); instead?
BTW I do not use camel cased or upper case names in the core.

Dietz0r: Moving it to planner.c as static inlined would make it very hard to find if you wanted to have other profiles instead of the generic placeholder 20% ones. 

I shied away from making it yet another list of $ settings because it would likely mean a whole slew of settings for however much profiles you wanted. having the lookuptable in settings.c makes it findable though. 

maybe move it to the driver my_machine.h because it is a compile time setting?

settings.h
Dietz0r: Would something like this work? 
this in gc_init() to initialize the state:
gc_init
and call back it in the limit function in planner.c like this:
planner c


and then set and reset the gc_state.modal.activeaccelprofile to whatever profile is currently needed with the g187 function?

settings.h
```diff
@@ -521,7 +521,11 @@ typedef enum {
     // Calculated base values for driver/plugin stepper settings
     Setting_AxisExtended0        = Setting_AxisSettingsBase2,
     Setting_AxisExtended1        = Setting_AxisSettingsBase2 + AXIS_SETTINGS_INCREMENT,
+#if ENABLE_JERK_ACCELERATION
+    Setting_AxisJerk             = Setting_AxisSettingsBase2 + 2 * AXIS_SETTINGS_INCREMENT,
+#else
```
terjeio, line 526: Setting_AxisExtended2 is used by the Trinamic code...
The extended axis settings are aimed at plugin code so should not be used by the core. I want to move Setting_AxisStepperCurrent and Setting_AxisMicroSteps to the extended settings in order to free them up for the core but I have not figured out how I can do that in a backward compatible way. So I am considering adding a new block of axis settings for core use. For now you can switch to Setting_AxisExtended8 instead?

settings.h
```diff
@@ -714,6 +718,9 @@ typedef struct {
     float steps_per_mm;
     float max_rate;
     float acceleration;
+#if ENABLE_JERK_ACCELERATION   
+    float jerk;
+#endif
```
terjeio, line 723: This is another potential problem we have to be aware of, the area reserved for global settings may overflow when many axes are configured. I have already switched to claim settings storage for a second PWM spindle from above the area reseved for the core. Perhaps the jerk settings should as well? I'll have to think about that. See nvs.h.

terjeio (commented):
I am starting to believe that this PR should be deferred until I get an overdue settings revision done. This so that I do not run into issues later on. I'll look into this in the next few days.

settings.h
Dietz0r: Will do. 
I just had it on the Setting_AxisExtended2 because it lines up with the normal Acceleration settings number, but there's no harm in moving it out of the way.

distebia (approved):

settings.h
terjeio: I have just [commited a version](https://github.com/grblHAL/core/blob/master/changelog.md#20241208) with updated settings structures and that adds a [permanent field for jerk](https://github.com/grblHAL/core/blob/a41790dada31aa0fc1c49b92fe4df033a9533079/settings.h#L715). I have also reserved ten new axis settings for the core, [jerk beeing the first assigned](https://github.com/grblHAL/core/blob/a41790dada31aa0fc1c49b92fe4df033a9533079/settings.h#L527). 

If you can update your PR against this version then I will be able to merge it without risking issues with the settings.

Dietz0r: Currently at work, will take a look and push latest revision tomorrow night or maybe wednesday! 
Thanks for all the work! :)

planner.c
```diff
@@ -348,9 +348,28 @@ static inline float limit_acceleration_by_axis_maximum (float *unit_vec)
         if (unit_vec[--idx] != 0.0f)  // Avoid divide by zero.
             limit_value = min(limit_value, fabsf(settings.axis[idx].acceleration / unit_vec[idx]));
     } while(idx);
+#if ENABLE_ACCELERATION_PROFILES
+    limit_value *= lookupprofile(gc_state.modal.activeaccelprofile);
```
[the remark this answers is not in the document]
Dietz0r: I rewrote it so it uses the plan_block_t structure and plan_data_init() function.

How it should now work is the machine initializes with active profile 1 - 1.0 factor. in gcode.c (including reset handling)
plan_data_init() looks up the current factor with the lookupfactor() function.
plan_block_t now has storage for the current_factor and gets it from pl_data->current_factor to calucalte adjusted acceleration limits via  limit_value *= block->acceleration_factor;

hope i didnt mess that up too badly

settings.h
Dietz0r: resolved the conflicts. :)

terjeio:
Is this nearly ready for merge now?
FYI settings.c is out of sync with the core, one removed function is added again (set_offset_lock) and at least one modified one is duplicated by adding the earlier code (set_homing_enable).

Dietz0r:
> Is this nearly ready for merge now? FYI settings.c is out of sync with the core, one removed function is added again (set_offset_lock) and at least one modified one is duplicated by adding the earlier code (set_homing_enable).

The out of sync file might have been from cleaning up the merge conflicts, my apologes.
I just did a test compile to make sure it works. Found one missing declaration. 
It does compile but i didnt have time yet to verify the new G187 code works as it should.

So apart from physical testing i guess it's ready?

Dietz0r:
Okay, seems to be looking good now on my end.

planner.c
```diff
@@ -339,7 +339,7 @@ inline static float plan_compute_profile_parameters (plan_block_t *block, float
     return nominal_speed;
 }
 
-static inline float limit_acceleration_by_axis_maximum (float *unit_vec)
+static inline float limit_acceleration_by_axis_maximum (float *unit_vec,)
 {
```
terjeio, line 343: Stray comma?

settings.h
```diff
@@ -712,7 +712,9 @@ typedef struct {
     float dual_axis_offset;
     float homing_seek_rate;
     float homing_feed_rate;
+#if ENABLE_JERK_ACCELERATION
```
terjeio, line 715: Remove this #if/#endif, it will trigger a settings reset when enabling/disabling jerk.

planner.c
Dietz0r: Yep, removed.

settings.h
Dietz0r: Done.

Dietz0r:
Also fixed the G187 Unused P Value error. The crash on motion command still persists. :/

terjeio:
No crashes for me, but then I have modified your code a bit, in _gcode.c_:

Simplified this a bit, the FAIL macro is a `return` statement so no code will be executed after.
And p < 1 is not neccesarily a negative value...
`gc_state.modal.acceleration_profile` holds the factor, should be renamed to match `plan_data.acceleration_factor`.
```
#if ENABLE_ACCELERATION_PROFILES

        case NonModal_SetAccelerationProfile:

            if(gc_block.words.e)
                FAIL(Status_GcodeUnsupportedCommand);

            if(gc_block.words.p && (gc_block.values.p < 1.0f || gc_block.values.p > 5.0f))
                FAIL(Status_GcodeValueOutOfRange);

            gc_state.modal.acceleration_profile = lookupfactor(gc_block.words.p ? (uint8_t)gc_block.values.p - 1 : 0);
            gc_block.words.p = Off;
            break;

#endif
```
Copying the factor to plan_data passed to the planner is missing in your code:
```
    // Initialize planner data struct for motion blocks.
    plan_line_data_t plan_data;
    memset(&plan_data, 0, sizeof(plan_line_data_t)); // Zero plan_data struct
    plan_data.offset_id = gc_state.offset_id;
    plan_data.condition.target_validated = plan_data.condition.target_valid = sys.soft_limits.mask == 0;
#if ENABLE_ACCELERATION_PROFILES
    plan_data.acceleration_factor = gc_state.modal.acceleration_profile;
#endif
```

Dietz0r:
image
tested and working ... this took a while <3 thanks for all your help

PR #603, opened 2024-10-09
Add experimental support for string registers

skasti:
Taking inspiration from named ngc-parameters, I have added experimental support for string registers. These can currently only be used to insert the text from the registers into messages to the sender `(MSG,`, `(DEBUG,` and `(PRINT,`, but that is also the use-case I view as most useful for now.

Example use is for storing tool descriptions;
``` gcode
(&1=T1: 4mm 3flute endmill)
(&2=T2: 6mm 45deg chamfer)
(&3=T3: 5.9mm 60deg thread mill)
```

upon tool change, you can give the user a description of the tool to load:
```gcode
(MSG: Please load &[#<_selected_tool>] and press 'cycle start')
M1; Wait for user to press cycle start
```

Or a little more complicated:
P210 (store tool information):
```gcode
(debug,Store tool info)
#<tool>=#20
#<type>=#17
#<diameter>=#7
#<custom>=#8
(debug,T#<tool> Q#<type> D#<diameter> E#<custom>)
; Start by incementing the index
#<_scnc_info_index>=[#<_scnc_info_index>+1]
(debug,Info-index: #<_scnc_info_index>)
; Then calculate the starting param for the tool info
#<info_offset>=[#<_scnc_info_index>*#<_scnc_info_size>]
#<info_start>=[#<_scnc_info_start>+#<info_offset>]
(debug,Info-start: #<info_start>)
#<_scnc_desc_index>=[#<_scnc_desc_index>+1]
(debug,Description-index: #<_scnc_desc_index>)
#<desc_register>=[#<_scnc_desc_start>+#<_scnc_desc_index>]
(debug,Description-register: #<desc_register>)
G65P3I[#<info_start>]Q[#<tool>]
G65P3I[#<info_start>+1]Q[#<type>]
G65P3I[#<info_start>+2]Q[#<diameter>]
G65P3I[#<info_start>+3]Q[#<custom>]
G65P3I[#<info_start>+4]Q[#<desc_register>]

(&[#<desc_register>]=T#<tool>: L#<length> D#<diameter>)
O60 if [#<custom> NE 0]
    (&[#<desc_register>]=&[#<desc_register>] &[#<_scnc_customtype>+#<type>]#<custom> &[#<type>])
O60 else
    (&[#<desc_register>]=&[#<desc_register>] &[#<type>])
O60 endif

#<_tool_desc>=#<desc_register>
```

P200.macro (find rack slot for tool):
```gcode
#<_tool>=#20

G65P201T[#<_tool>]Q[#17]D[#7]E[#8]

G65P111R[#<_tool>]
o70 if [#<_pickupSlot> GE 0]
    #<slot>=[#<_pickupSlot> + 1]
    (MSG,Tool T#<_tool> already stored in slot #<slot> from the back)
    M99
o70 endif

G65P110R[#<_tool>]
o80 if [#<_parkSlot> GE 0]
    #<slot>=[#<_parkSlot> + 1]
    (MSG,Put tool &[#<_tool_desc>] in slot #<slot> from the back)
    M0
    G65P112Q[#<_parkSlot>]R[#<_tool>]
o80 else
    (MSG,Unable to find slot for tool &[#<_tool_desc>]. It will have to be manually loaded/unloaded)
    M0
o80 endif
M99
```

Then in my programs, I list the tools at the start:
```gcode
G65P200T1Q4D4E10
G65P200T4Q7D6E0.5
```

which then results in messages for the operator, so that they understand which tools to load where:
`Put tool T1: D4 L10 flat end mill in slot 1 from the back`
`Put tool T4: D6 CR0.5 bull nose end mill in slot 2 from the back`

Updated to reflect how setting string registers is done using comments

### DIFF
--- a/config.h
+++ b/config.h
@@ -534,7 +534,7 @@ Set to \ref On or 1 to enable experimental support for parameters.
 Set to \ref On or 1 to enable experimental support for string registers.
 */
 #if !defined STRING_REGISTERS_ENABLE || defined __DOXYGEN__
-#define STRING_REGISTERS_ENABLE On
+#define STRING_REGISTERS_ENABLE Off
 #endif
 
 /*! \def NGC_N_ASSIGN_PARAMETERS_PER_BLOCK
--- a/config.h
+++ b/config.h
@@ -529,12 +529,12 @@ Set to \ref On or 1 to enable experimental support for parameters.
 #define NGC_PARAMETERS_ENABLE On
 #endif
 
-/*! \def NGC_STRING_REGISTERS_ENABLE
+/*! \def STRING_REGISTERS_ENABLE
 \brief
 Set to \ref On or 1 to enable experimental support for string registers.
 */
-#if !defined NGC_STRING_REGISTERS_ENABLE || defined __DOXYGEN__
-#define NGC_STRING_REGISTERS_ENABLE On
+#if !defined STRING_REGISTERS_ENABLE || defined __DOXYGEN__
+#define STRING_REGISTERS_ENABLE On
 #endif
 
 /*! \def NGC_N_ASSIGN_PARAMETERS_PER_BLOCK
--- a/config.h
+++ b/config.h
@@ -529,6 +529,14 @@ Set to \ref On or 1 to enable experimental support for parameters.
 #define NGC_PARAMETERS_ENABLE On
 #endif
 
+/*! \def NGC_STRING_REGISTERS_ENABLE
+\brief
+Set to \ref On or 1 to enable experimental support for string registers.
+*/
+#if !defined NGC_STRING_REGISTERS_ENABLE || defined __DOXYGEN__
+#define NGC_STRING_REGISTERS_ENABLE On
+#endif
+
 /*! \def NGC_N_ASSIGN_PARAMETERS_PER_BLOCK
 \brief
 Maximum number of parameters allowed in a block.
--- a/gcode.c
+++ b/gcode.c
@@ -940,8 +940,14 @@ status_code_t gc_execute_block (char *block)
             if(gc_state.skip_blocks)
                 return Status_OK;
             float register_id;
-            if (!read_float(block, &char_counter, &register_id)) {
-                FAIL(Status_BadNumberFormat);   // [Expected register id]
+            if (block[char_counter] == '[') {
+                if (!ngc_eval_expression(block, &char_counter, &register_id)) {
+                    FAIL(Status_ExpressionSyntaxError);   // [Invalid expression syntax]
+                }
+            } else {
+                if (!read_float(block, &char_counter, &register_id)) {
+                    FAIL(Status_BadNumberFormat);   // [Expected register id]
+                }
             }
 
             if (block[char_counter++] != '=') {
--- a/gcode.c
+++ b/gcode.c
@@ -43,6 +43,10 @@
 #endif
 #endif
 
+#if STRING_REGISTERS_ENABLE
+#include "string_registers.h"
+#endif
+
 // NOTE: Max line number is defined by the g-code standard to be 99999. It seems to be an
 // arbitrary value, and some GUIs may require more. So we increased it based on a max safe
 // value when converting a float (7.2 digit precision) to an integer.
@@ -573,6 +577,14 @@ char *gc_normalize_block (char *block, char **message)
     if(*block == '/')
         block++;
 
+#if STRING_REGISTERS_ENABLE
+    // If the block starts with '@' it means it is setting a string register value,
+    // and we should not normalize it
+    if(*block == '@') {
+        return block;
+    }
+#endif
+
     s1 = s2 = block;
 
     while((c = *s1) != '\0') {
@@ -923,8 +935,29 @@ status_code_t gc_execute_block (char *block)
                 FAIL(Status_GcodeUnsupportedCommand); // [For now...]
 #endif
             }
-        }
+#if STRING_REGISTERS_ENABLE
+        } else if (letter == '@') {
+            if(gc_state.skip_blocks)
+                return Status_OK;
+            float register_id;
+            if (!read_float(block, &char_counter, &register_id)) {
+                FAIL(Status_BadNumberFormat);   // [Expected register id]
+            }
 
+            if (block[char_counter++] != '=') {
+                FAIL(Status_BadNumberFormat);   // [Expected equal sign]
+            }
+
+            if (!string_register_set((string_register_id_t)register_id, &block[char_counter++])) {
+                FAIL(Status_ExpressionInvalidArgument); // [Invalid value after '=']
+            }
+
+            // setting a string-register consumes the rest of this block
+            break;
+        }
+#else
+        }
+#endif
         if((gc_block.words.mask & o_label.mask) && (gc_block.words.mask & ~o_label.mask) == 0) {
             char_counter--;
             return ngc_flowctrl(gc_block.values.o, block, &char_counter, &gc_state.skip_blocks);
--- a/gcode.c
+++ b/gcode.c
@@ -941,7 +941,7 @@ status_code_t gc_execute_block (char *block)
                 return Status_OK;
             float register_id;
             if (block[char_counter] == '[') {
-                if (!ngc_eval_expression(block, &char_counter, &register_id)) {
+                if (ngc_eval_expression(block, &char_counter, &register_id) != Status_OK) {
                     FAIL(Status_ExpressionSyntaxError);   // [Invalid expression syntax]
                 }
             } else {
--- a/gcode.c
+++ b/gcode.c
@@ -944,10 +944,8 @@ status_code_t gc_execute_block (char *block)
                 if (ngc_eval_expression(block, &char_counter, &register_id) != Status_OK) {
                     FAIL(Status_ExpressionSyntaxError);   // [Invalid expression syntax]
                 }
-            } else {
-                if (!read_float(block, &char_counter, &register_id)) {
-                    FAIL(Status_BadNumberFormat);   // [Expected register id]
-                }
+            } else if (!read_float(block, &char_counter, &register_id)) {
+                FAIL(Status_BadNumberFormat);   // [Expected register id]
             }
 
             if (block[char_counter++] != '=') {
--- a/gcode.c
+++ b/gcode.c
@@ -352,6 +352,10 @@ void gc_init (void)
 
 #if NGC_EXPRESSIONS_ENABLE
     ngc_flowctrl_init();
+    #if STRING_REGISTERS_ENABLE
+    string_registers_init();
+    #endif
+    ngc_expr_init();
 #endif
 #if NGC_PARAMETERS_ENABLE
     ngc_modal_state_invalidate();
@@ -577,14 +581,6 @@ char *gc_normalize_block (char *block, status_code_t *status, char **message)
     if(*block == '/')
         block++;
 
-#if STRING_REGISTERS_ENABLE
-    // If the block starts with '&' it means it is setting a string register value,
-    // and we should not normalize it
-    if(*block == '&') {
-        return block;
-    }
-#endif
-
     s1 = s2 = block;
 
     while((c = *s1) != '\0') {
@@ -611,14 +607,16 @@ char *gc_normalize_block (char *block, status_code_t *status, char **message)
                         if(message && *message == NULL) {
 #if NGC_EXPRESSIONS_ENABLE
                             if(!strncasecmp(comment, "DEBUG,", 6)) { // Debug message string substitution
-                                if(settings.flags.ngc_debug_out) {
+                                if(settings.flags.ngc_debug_out && grbl.on_string_substitution) {
                                     comment += 6;
-                                    ngc_substitute_parameters(comment, message);
+                                    grbl.on_string_substitution(comment, message);
                                 }
                                 *comment = '\0'; // Do not generate grbl.on_gcode_comment event!
                             } else if(!strncasecmp(comment, "PRINT,", 6)) { // Print message string substitution
-                                comment += 6;
-                                ngc_substitute_parameters(comment, message);
+                                if(grbl.on_string_substitution) {
+                                    comment += 6;
+                                    grbl.on_string_substitution(comment, message);
+                                }
                                 *comment = '\0'; // Do not generate grbl.on_gcode_comment event!
                             } else {
 #endif
@@ -936,40 +934,8 @@ status_code_t gc_execute_block (char *block)
                 FAIL(Status_GcodeUnsupportedCommand); // [For now...]
 #endif
             }
-#if STRING_REGISTERS_ENABLE
-        } else if (letter == '&') {
-            if(gc_state.skip_blocks)
-                return Status_OK;
-            float register_id;
-            if (block[char_counter] == '[') {
-                if (ngc_eval_expression(block, &char_counter, &register_id) != Status_OK) {
-                    FAIL(Status_ExpressionSyntaxError);   // [Invalid expression syntax]
-                }
-            } else if (!read_float(block, &char_counter, &register_id)) {
-                FAIL(Status_BadNumberFormat);   // [Expected register id]
-            }
-
-            if (block[char_counter++] != '=') {
-                FAIL(Status_BadNumberFormat);   // [Expected equal sign]
-            }
-
-            char *strValue;
-            substitute_parameters(&block[char_counter++], &strValue);
-            report_message(strValue, Message_Debug);
-
-            if (!string_register_set((string_register_id_t)register_id, strValue)) {
-                free(strValue);
-                FAIL(Status_ExpressionInvalidArgument); // [Invalid value after '=']
-            } else {
-                free(strValue);
-            }
-
-            // setting a string-register consumes the rest of this block
-            break;
         }
-#else
-        }
-#endif
+
         if((gc_block.words.mask & o_label.mask) && (gc_block.words.mask & ~o_label.mask) == 0) {
             char_counter--;
             return ngc_flowctrl(gc_block.values.o, block, &char_counter, &gc_state.skip_blocks);
--- a/gcode.c
+++ b/gcode.c
@@ -578,9 +578,9 @@ char *gc_normalize_block (char *block, char **message)
         block++;
 
 #if STRING_REGISTERS_ENABLE
-    // If the block starts with '@' it means it is setting a string register value,
+    // If the block starts with '&' it means it is setting a string register value,
     // and we should not normalize it
-    if(*block == '@') {
+    if(*block == '&') {
         return block;
     }
 #endif
@@ -936,7 +936,7 @@ status_code_t gc_execute_block (char *block)
 #endif
             }
 #if STRING_REGISTERS_ENABLE
-        } else if (letter == '@') {
+        } else if (letter == '&') {
             if(gc_state.skip_blocks)
                 return Status_OK;
             float register_id;
--- a/gcode.c
+++ b/gcode.c
@@ -561,7 +561,7 @@ bool gc_modal_state_restore (gc_modal_t *copy)
 // If the driver handles message comments then the first is extracted and returned in a dynamically
 // allocated memory block, the caller must free this after the message has been processed.
 
-char *gc_normalize_block (char *block, char **message)
+char *gc_normalize_block (char *block, status_code_t *status, char **message)
 {
     char c, *s1, *s2, *comment = NULL;
 
@@ -620,12 +620,8 @@ char *gc_normalize_block (char *block, char **message)
                                 comment += 6;
                                 ngc_substitute_parameters(comment, message);
                                 *comment = '\0'; // Do not generate grbl.on_gcode_comment event!
-                            } else if(!strncasecmp(comment, "MSG,", 4)) {
-                                    comment += 4;
-                                    ngc_substitute_parameters(comment, message);
-                                }
-                            }
-#else
+                            } else {
+#endif
                             size_t len = s1 - comment - 3;
 
                             if(!strncasecmp(comment, "MSG,", 4) && (*message = malloc(len))) {
@@ -637,12 +633,13 @@ char *gc_normalize_block (char *block, char **message)
                                 }
                                 memcpy(*message, comment, len);
                             }
-                        }
+#if NGC_EXPRESSIONS_ENABLE
+                            }
 #endif
+                        }
                     }
-
                     if(*comment && *message == NULL && grbl.on_gcode_comment)
-                        grbl.on_gcode_comment(comment);
+                        *status = grbl.on_gcode_comment(comment);
                 }
                 comment = NULL;
                 break;
@@ -785,6 +782,7 @@ status_code_t gc_execute_block (char *block)
 #endif
 
     char *message = NULL;
+    status_code_t status = Status_OK;
     struct {
         float f;
         uint32_t o;
@@ -792,12 +790,15 @@ status_code_t gc_execute_block (char *block)
         tool_id_t t;
     } single_meaning_value = {0};
 
-    block = gc_normalize_block(block, &message);
+    block = gc_normalize_block(block, &status, &message);
+
+    if(status != Status_OK)
+        FAIL(status);
 
     if(block[0] == '\0') {
         if(message)
             gc_output_message(message);
-        return Status_OK;
+        return status;
     }
 
     // Determine if the line is a program start/end marker.
--- a/gcode.c
+++ b/gcode.c
@@ -952,8 +952,15 @@ status_code_t gc_execute_block (char *block)
                 FAIL(Status_BadNumberFormat);   // [Expected equal sign]
             }
 
-            if (!string_register_set((string_register_id_t)register_id, &block[char_counter++])) {
+            char *strValue;
+            substitute_parameters(&block[char_counter++], &strValue);
+            report_message(strValue, Message_Debug);
+
+            if (!string_register_set((string_register_id_t)register_id, strValue)) {
+                free(strValue);
                 FAIL(Status_ExpressionInvalidArgument); // [Invalid value after '=']
+            } else {
+                free(strValue);
             }
 
             // setting a string-register consumes the rest of this block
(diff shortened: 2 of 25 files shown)


config.h
```diff
@@ -529,6 +529,14 @@ Set to \ref On or 1 to enable experimental support for parameters.
...
 #endif
 
+/*! \def STRING_REGISTERS_ENABLE
+\brief
+Set to \ref On or 1 to enable experimental support for string registers.
+*/
+#if !defined STRING_REGISTERS_ENABLE || defined __DOXYGEN__
+#define STRING_REGISTERS_ENABLE On
+#endif
+
```
skasti, line 539: I must admit I am not 100% sure I understand how this works, but it did seem to work with PIO atleast 😅

gcode.c
```diff
@@ -591,7 +595,7 @@ static void substitute_parameters (char *comment, char **message)
 {
     size_t len = 0;
     float value;
-    char *s, c;
+    char *s, c, *strValue;
```
skasti, line 598: should I put this separately within a `#if STRING_REGISTERS_ENABLE`-block?

gcode.c
```diff
@@ -1060,8 +1100,29 @@ status_code_t gc_execute_block (char *block)
             }
 
             continue;
-        }
+#if STRING_REGISTERS_ENABLE
+        } else if (letter == '@') {
+            if(gc_state.skip_blocks)
+                return Status_OK;
+            float register_id;
+            if (!read_float(block, &char_counter, &register_id)) {
```
skasti, line 1108: ah, this limits the setting of a string register to exact values. Should use read_param here 🤔

skasti: Fixed

skasti:
Converted to draft as I got some bugs when I want to use parameters in strings 😅

gcode.c
```diff
@@ -726,6 +758,14 @@ char *gc_normalize_block (char *block, char **message)
     if(*block == '/')
         block++;
 
+#if STRING_REGISTERS_ENABLE
+    // If the block starts with '@' it means it is setting a string register value,
+    // and we should not normalize it
+    if(*block == '@') {
+        return block;
+    }
+#endif
```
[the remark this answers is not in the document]
skasti: One thing I was considering is to use quotes for strings, and do something similar to comments, where characters between quotes are not modified.

So instead of `@5=This is a test`, you would write `@5="This is a test"`. This might help with one issue I had with expressions, where putting a space between values and operators causes an invalid expression error ( so `@[#<someParam> + 5]=Test` gives error, but `@[#<someParam>+5]=Test` does not. In my opinion spaces should just be ignored, and I think that is done by the block normalization? 🤔

terjeio:
FYI `@` and `^` are used by LinuxCNC for [polar coordinates](http://linuxcnc.org/docs/stable/html/tr/gcode/overview.html#gcode:polar-coordinates) which I plan to add support for later...
I'll take a closer look at this when I am back at base next week.

skasti:
> FYI `@` and `^` are used by LinuxCNC for [polar coordinates](http://linuxcnc.org/docs/stable/html/tr/gcode/overview.html#gcode:polar-coordinates) which I plan to add support for later... I'll take a closer look at this when I am back at base next week.

Hah, damnit. I can't very well take `@` when it has a well-defined usage already 😅🙈 But it would be very elegant to have a special character for string registers, since using `SR` or similar makes it easy to mix up with normal text. Will see if some other character might suffice 🤔
Maybe paragraph (`§`), since that hints at text even more than @ 🤔

skasti:
Looks like ampesand (`&`) works, and I did not find any uses for it in grblHAL or gcode that this would interfere with 🤔

config.h
skasti: Have read up a bit, and now see that this makes it enabled by default, so I should change this 😅

config.h
```diff
@@ -529,6 +529,14 @@ Set to \ref On or 1 to enable experimental support for parameters.
 #define NGC_PARAMETERS_ENABLE On
 #endif
 
+/*! \def STRING_REGISTERS_ENABLE
+\brief
+Set to \ref On or 1 to enable experimental support for string registers.
+*/
+#if !defined STRING_REGISTERS_ENABLE || defined __DOXYGEN__
+#define STRING_REGISTERS_ENABLE On
```
skasti, line 537:
```suggestion
#define STRING_REGISTERS_ENABLE Off
```

terjeio:
I've been pondering over this while going back and improving the whole parameters and expression implementation...
While doing so I got a new idea for setting string registers, by using comments and trapping them via a `grbl.on_gcode_comment` event handler.
E.g: `(SR[<expr>]=some string)`
This is a lot cleaner and solves the issue of block normalization upper casing the string argument. Outputting strings could be achieved in a similar manner?

terjeio:
Something like this to set string registers:
```
static void onGcodeComment(char *comment)
{
    int32_t sr;
    uint_fast8_t pos = 3;

    if(!strncmp(comment, "(SR[", 4)) {
        if((ngc_read_integer_value(comment, &pos, &sr)) == Status_OK) {
// check for negative index...
            if(comment[pos] == '=') {
                char c, buf[MAX_SR_LENGTH + 1], *bptr = buf;

                while((c = comment[++pos])) {
                    *bptr++ = c;
                    // perform substitution (expose gcode.c function?)
                }
                *bptr = '\0';
                string_register_set((string_register_id_t)sr, buf);
            }
        }
    }
// return status code
}

void my_plugin_init (void)
{
    grbl.on_gcode_comment = onGcodeComment;
}
```

`grbl.on_gcode_comment` should be changed to return a status code, and your `&` syntax to read values within comments should remain? At least string registers can then live entirely within comments which is, IMO, preferable.

skasti:
> I've been pondering over this while going back and improving the whole parameters and expression implementation... While doing so I got a new idea for setting string registers, by using comments and trapping them via a `grbl.on_gcode_comment` event handler. E.g: `(SR[<expr>]=some string)` This is a lot cleaner and solves the issue of block normalization upper casing the string argument. Outputting strings could be achieved in a similar manner?

While it would work, I am not sure I like it. Having comments that are altering semi-persistent stuff feels a bit "wrong", even though I understand the sentiment 😅 And although the current use for string-registers is strictly in comments, I would not be surprised if someone would use them to create engraving plugins etc, and then it would start to look strange to set/modify them only within comments.

As for solving the normalization, I think using quotes will be an elegant solution, as it would then also work if someone at some point adds some other feature that needs to use strings for anything else 🤔 And parameter_substitution can be done in the normalization, which would make that feature available for any feature that needs strings in the future.

But in any case, I would probably keep the syntax for setting and reading similar;
```
(&10=some text)
(MSG,Here is &10)
```
or
```
(SR[10]=some text)
(MSG,Here is SR[10])
```
not

```
(SR[10]=some text)
(MSG,Here is &10)
```

🤔

skasti:
This belongs in a discussion, but one thing I have thought a little about somehow making a "handler-system" where you would have the default handlers in grbl-core for everything that is core-functionality (supported G-codes and M-codes etc), and plugins could register either customizations of existing handlers or add completely new handlers. That way this could be a plugin that registered a "block-handler" and a "substitution-handler" for `&` that would handle the setting of and substitution of string registers.

Though I am far away from understanding enough of the code to be able to implement something like this yet other than that it would require a complete rewrite of most of gcode-handling, and is borderline impossible without a test-framework 😅

terjeio:
> ... and then it would start to look strange to set/modify them only within comments.

Strings outside comments in G-code is even stranger, no dialects I know of supports them... The Fanuc SR registers are for robots which is a very specialiced segment? There is also prior use of sending data outside the scope of normal G-code via comments, IIRC LinuxCNC passes some info to its plasma THC implementation that way. And if IIRC that is the reason I added the `grbl.on_gcode_comment` event (which by the way does not strip  the leading `(` character as it should - to be fixed...

I agree that _gcode.c_ would benefit from a rewrite, it is an insanely long function. But it is not something that is achieved in a few afternoons, and it works, so better to leave it for later when there is nothing else to improve on?

I have been tinkering a bit more with your string support and want to use it to implement named subroutine calls: `o<subname> call ...` This by limiting the number of string registers to 65535 (16 bit) and auto-assign an id to named subroutines from max 32 bit value counting down. This way I can pass the subroutine name by id...

A tip: the latest versions supports indexed parameter access with ##n and #[#n] syntax (n can be an expression].

string_registers.c
```diff
@@ -0,0 +1,119 @@
...
+string_register_t *find_string_register_with_last (string_register_id_t id, string_register_t **last_register) {
+    string_register_t *current_register = string_registers;
+    int i = 0;
+
+    while(current_register != NULL) {
+
+        if(current_register->id == id) {
+            return current_register;
+
+        } else if (i++ < 1000) {
```
terjeio, line 57: This is an arbitrary limitation that I see no reason to have.

string_registers.c
```diff
@@ -0,0 +1,119 @@
...
+
+#include "system.h"
+#include "settings.h"
+#include "ngc_params.h"
+#include "string_registers.h"
+
+#ifndef MAX_SR_LENGTH
+#define MAX_SR_LENGTH 50
+#endif
+typedef struct string_register {
```
terjeio, line 40: If you move the value array to the end and change it to a single character (value[1]) and malloc the struct size plus the string length then you may save memory and lift (or increase) the maximum string length. Max. block (line) length in gcode is 256, this limit should perhaps be adhered to - remember plugin code may use the API for string handling directly.

string_registers.c
skasti: this is/was a "safety-feature" I added as I was having some issue where the while-loop got stuck forever 😅 Should be good to remove now 👍

string_registers.c
skasti: Wouldn't malloc'ing dynamicly like that cause issues if you modify a string-register with a longer value than it was originally malloc'ed with?

skasti:
Been busy "playing" carpenter this weekend, so have to try to rebase and add the discussed changes some time during the coming week 👍

string_registers.c
terjeio: Use realloc instead and check for increased length.

skasti: huh, I thought an array of structs in C would be lined up one after another in memory, so I assumed each string_register had to be the same size 🤔 😅

terjeio: an array yes, but not a linked list - the last element in the struct can be any length since the struct is dynamically allocated and the next pointer points to the next element wherever that has been allocated...

skasti: lol, of course you are correct. I somehow skipped over the fact that it is a linked list and not an array! 🤦 😂

terjeio: > I have to update the pointer to "next" in the previous item when I realloc an existing register, right?

Only if realloc cannot extend the current register allocation. But easier to just assume it has been moved.
And there is no need to use malloc at all for the initial allocation, if you pass a null pointer to realloc it behaves like malloc.

terjeio:
[Here is a link](http://linuxcnc.org/docs/2.8/html/plasma/plasmac-user-guide.html#magic-comments) to prior use of comments to pass non-standard information.
Perhaps it would be useful to have two options to set string registers, one that does parameter substitution and one that does not, leaving the option to perform substitution later?
FYI in the next commit I'll change the signature of `grbl.on_gcode_comment` to add status code returns:

```
static status_code_t onGcodeComment(char *comment)
{
    int32_t sr;
    uint_fast8_t pos = 2;
    status_code_t status = Status_OK;

    if(!strncasecmp(comment, "SR[", 3)) { // or if(!strncasecmp(comment, "&[", 2)) {
        if((status = ngc_read_integer_value(comment, &pos, &sr)) == Status_OK) {
            if(comment[pos] == '=') {
                char *buf = NULL;
                if(ngc_substitute_parameters(comment + ++pos, &buf)) {
                    if(!string_register_set((string_register_id_t)sr, buf))
                        status = Status_FlowControlOutOfMemory;
                    free(buf);
                } else
                        status = Status_FlowControlOutOfMemory;
            } else
                status = Status_ExpressionSyntaxError;
        }
    } else if(on_gcode_comment)
        status = on_gcode_comment(comment);

    return status;
}
```

string_registers.c
```diff
@@ -0,0 +1,123 @@
...
+}
+
+bool string_register_exists (string_register_id_t id) {
+    return find_string_register(id) != NULL;
+}
+
+bool string_register_set (ngc_param_id_t id, char *value) {
+    size_t length = strlen(value);
+    if (length > MAX_SR_LENGTH) {
+        report_message("String register values cannot be longer than 40 characters", Message_Warning);
```
terjeio, line 88: messages should not be output here. It should be done by the caller, if needed, by checking the return value.

terjeio:
[Another comment link](http://linuxcnc.org/docs/2.8/html/remap/remap.html#_error_handling_failing_a_remapped_code_ngc_procedure) that is interesting. I did not know about it and perhaps it should be supported?

string_registers.c
skasti: Ah, yes. I should probably return something more useful than a bool, then 😅

skasti:
> Perhaps it would be useful to have two options to set string registers, one that does parameter substitution and one that does not, leaving the option to perform substitution later? 

Hm, I haven't thought about that. It would be a bit more work to do double-substitution when fetching a string-register, but I guess it might be useful 🤔 I have been using substitution when setting string register to append information to existing tool-description text for now, and was very happy about how that works 🤔 
```gcode
&[#<desc_register>]=T#<tool>: L#<length> D#<diameter>
O60 if [#<custom> NE 0]
    &[#<desc_register>]=&[#<desc_register>] &[#<_scnc_customtype>+#<type>]#<custom> &[#<type>]
O60 else
    &[#<desc_register>]=&[#<desc_register>] &[#<type>]
O60 endif
```

Haven't gotten as far as to rewrite to setting registers within comments yet. Maybe tonight 😅 
Also have not had time to compile and test on my machine, so I would not take it for granted that this works right now, with the refactoring etc :P

skasti:
@terjeio Trying to move setting string registers to a onGcodeComment-function in string-registers.c causes a circular reference because getting the id of the register to set uses ngc_eval_expression, but ngc_substitute_parameters does string-register substitution and so needs to know about string registers as well 🤔 😅 

Maybe substitution of parameters and string registers also should be done using a core handler, like onGcodeComment? 🤔

skasti:
I have now done some initial testing, and it does seem like this is working. But I have not yet figured out a good way to check for memory leaks. How do you do this, @terjeio ? 🤔 

Would be nice if it was possible to get some type of unit testing on the repo as well, so it would be possible to test without deploying to a controller. Not sure how to do that though 😞

terjeio:
> Trying to move setting string registers to a onGcodeComment-function in string-registers.c causes a circular reference because getting the id of the register to set uses ngc_eval_expression, but ngc_substitute_parameters does string-register substitution and so needs to know about string registers as well

True - but you do not need to include the .h file, just declare the needed function, string_register_get(), as extern in _ngc_expr.c_. The way you have done it now is, IMO, overly complicated.

I am not keen on the new custom status code returns you have added, the old signatures were just fine - just do not output any messages in the string registers code, leave it to the caller if it is needed. BTW string substitution in comments should never cause an error return to the top level - any fault should be handled gracefully. If you insists on having status code returns add them to error.h if there is no suitable codes available there already.

> But I have not yet figured out a good way to check for memory leaks. How do you do this

By testing edge cases, and by knowing the behaviour of library calls. Still I miss some sometimes...
In fact you have a potential memory leak that I spotted, if realloc fails (returns null) any previously allocated memory has to be freed in your code.

> Would be nice if it was possible to get some type of unit testing on the repo as well, so it would be possible to test without deploying to a controller. Not sure how to do that though

You can use the [Simulator driver](https://github.com/grblHAL/Simulator). I prefer to run tests on a controller, and I make heavy use of the debugger while doing so.

I want to add some functions your string register code in order to support named o subroutines/calls via some trickery (by using 32 bit ids).  To do this I want you to use `ngc_param_id_t` instead of `string_register_id_t` in your functions, and I'll change `string_register_id_t` to 32 bit and use it in my "extended" functions. Ok?

skasti:
> By testing edge cases, and by knowing the behaviour of library calls. Still I miss some sometimes... In fact you have a potential memory leak that I spotted, if realloc fails (returns null) any previously allocated memory has to be freed in your code.

Hm, why, though? if realloc returns null, my intention was that the string register just remains unchanged? 🤔

skasti:
> > my intention was that the string register just remains unchanged
> 
> Isn't that bad behaviour? I would remove it from the list and return false indicating that setting the register failed.

Not really. Take my example where I append more content on a string register; if appending the extra info fails, I would much prefer to have the base data so that I can try again, rather than losing everything 😅 Regardless; it is not a memory leak when it can still be reassigned later, right? 🤔

skasti:
> True - but you do not need to include the .h file, just declare the needed function, string_register_get(), as extern in _ngc_expr.c_. The way you have done it now is, IMO, overly complicated.

Ah, I forgot that you could do that in C/C++ 😅 While my solution is more complex, it does separate some responsibility.. (got to catch a train...)

terjeio:
> I would much prefer to have the base data so that I can try again, rather than losing everything

I would rather not have som random data coming back later if I decide to reuse a register...
Recently I spent quite a lot of time on a somewhat similar issue with a STM32 processor where memory contents I changed were replaced with the original value after a soft reset due to a processor bug. Not funny...

skasti:
> > I would much prefer to have the base data so that I can try again, rather than losing everything
> 
> I would rather not have som random data coming back later if I decide to reuse a register... Recently I spent quite a lot of time on a somewhat similar issue with a STM32 processor where memory contents I changed were replaced with the original value after a soft reset due to a processor bug. Not funny...

With all due respect, that is a completely unrelated issue (persisted vs working memory). It is not really that surprising that a change made during runtime is not persisted after a soft reset if there is some issue that causes persistence to fail.

Also, I have not checked, but I am pretty sure that if you change a setting using $x= or a parameter using #x= and it fails, it does not erase the previous value, as that would be very unexpected and destructive behaviour.

Any operation should either have the desired effect, or no effect. This is just best practice in software development.

skasti:
> I am not keen on the new custom status code returns you have added, the old signatures were just fine - just do not output any messages in the string registers code, leave it to the caller if it is needed. BTW string substitution in comments should never cause an error return to the top level - any fault should be handled gracefully. If you insists on having status code returns add them to error.h if there is no suitable codes available there already.

Hm, I agree that string substitution should handle errors more gracefully and not end up returning a error code. And maybe these messages that I posted were mostly not that necessary outside me trying to debug the feature itself 🤔 

Though I guess returning an error code for "input too long/large" would be good, so I can add that if none exists already, maybe? 🤔

skasti:
> I want to add some functions your string register code in order to support named o subroutines/calls via some trickery (by using 32 bit ids). To do this I want you to use `ngc_param_id_t` instead of `string_register_id_t` in your functions, and I'll change `string_register_id_t` to 32 bit and use it in my "extended" functions. Ok?

I am not sure I understand why this would require using `ngc_param_id_t` to fetch string registers? I am guessing you would be using named parameters for the named subroutines, and store the subroutine itself in a string register? if so, would you not store the string register id in the value of the named param? 🤔

terjeio:
> Also, it would not be random data, it would be whatever it was before the failed attempt. 

That depends on how you view it. If I run a gcode program that sets a string register and later I run a new one that sets the same register and fails then I would consider it kind of random, or at least totally unrelated, to what I want it to be.

> I am not sure I understand why this would require using ngc_param_id_t to fetch string registers?

There are two variants of o call and o sub, [numbered](https://www.linuxcnc.org/docs/html/gcode/o-code.html#ocode:subroutines) where the subroutine is part of the running gcode program and [named](https://www.linuxcnc.org/docs/html/gcode/o-code.html#ocode:calling-files) which is another gcode program located in a filing system where name = the filename of the file. I have already added support for numbered subroutines, to add support for named ones I want to use string registers for temprary storage of the name and use the (automatically assigned) string register id to pass information to ngc_expr.c. The reason for this is primarily that gcode is not designed to handle strings. And since ngc_param_id_t is 16 bit I can differentiate between a "normal" string register and one used by ngc_expr.c for subroutines by assigning ids above 2^16.  Some inactive code [here](https://github.com/grblHAL/core/blob/1444c08997827a5653eea821f4739fdc8a689a78/gcode.c#L915-L923) and [here](https://github.com/grblHAL/core/blob/1444c08997827a5653eea821f4739fdc8a689a78/ngc_flowctrl.c#L638-L660) if you want to take a look.

skasti:
> That depends on how you view it. If I run a gcode program that sets a string register and later I run a new one that sets the same register and fails then I would consider it kind of random, or at least totally unrelated, to what I want it to be.

I would argue that in many scenarios, you would want the value to remain unchanged. IE; you have a string register with the current serial number as a string. every time you run your engraving code, you increment the serial number and regenerate the text. If unable to do so, I would prefer to fetch the register and see what the last successful value was, rather than it being deleted.

In any case; Your gcode program will stop and alert you if there is an error writing to a string register, but it is a lot more hassle to make it repair destroyed data. You would have to make a copy of the register you are about to change first (and a parameter to keep track of the id of the register), and then you have to have a separate program to restore the register that you could run after the program failed 😅 🙈 


I would probably use a named parameter to store the string-register id of the subroutine, and then use the string register to look up the file name. no need to mix the Ids? probably missing something here, as I am at work. Will read through the code later and see if I understand then 😅

skasti:
@terjeio Reading through [Calling Files](https://www.linuxcnc.org/docs/html/gcode/o-code.html#ocode:calling-files), I get the feeling it would be cleaner to add `gc_block.values.o_string` as a `char*`, and add the handler `grbl.on_named_macro_execute(char* name)`  so that it can be implemented the same way as macros? 🤔

skasti:
Ok, I see there is much here I still do not understand. I thought this was only going to be used for a single block (`O<mysubroutine> CALL`), and that this meant it was OK for it to just be viable for that one block 😅🙈
So I will not protest against using string registers for this, though I guess this means I should make a way to delete a string register, so that temporary registers can be cleaned up? 🤔 

But I still do not fully understand why you want string-registers to use `ngc_param_id_t` as id. string-registers are not params 😅 
If you want to use the mechanism you mention, it would be "just" as easy to make the `string_register_set_name`-function return a 32bit uint that is `NGC_MAX_PARAM_ID` plus the actual string register id, and then subtract `NGC_MAX_PARAM_ID` from `o_label` again` in `ngc_flowctrl`?

gcode.c
```diff
@@ -599,14 +607,16 @@ char *gc_normalize_block (char *block, status_code_t *status, char **message)
                         if(message && *message == NULL) {
 #if NGC_EXPRESSIONS_ENABLE
```
skasti, line 608: @terjeio Should handling of `(DEBUG,` and `(PRINT,` be moved to `ngc_flowctrl`'s `onGcodeComment`? 🤔

terjeio: No. Since substitution is done in _ngc_expr.c_ it belongs there, and also because the` grbl.on_gcode_comment()` event is not able to return a string (and should not be changed to accomodate that).

Your `grbl.on_string_substitution()` event could be set up at startup to point to a new function in _ngc_expr.c_, making it overridable. And then your string register handling could (or rather should?) be moved to a plugin, in [this repo](https://github.com/grblHAL/Plugins_misc/)? `grbl.on_string_substitution`  has to be changed to a more descriptive name if doing so...

skasti: Aha, maybe I should make a separate PR with just the string substitution-stuff, and move string registers to a plugin-repo? 🤔

terjeio: >            case ')':
                if(comment && !gc_state.skip_blocks) {
                    *s1 = '\0';
                    if(!hal.driver_cap.no_gcode_message_handling) {

                        if(message && *message == NULL) {

                            if(grbl.on_process_gcode_comment)
                                *message = grbl.on_process_gcode_comment(comment);

                            if(*message == NULL) {
                                size_t len = s1 - comment - 3;

                                if(!strncasecmp(comment, "MSG,", 4) && (*message = malloc(len))) {

                                    comment += 4;
                                    while(*comment == ' ') {
                                        comment++;
                                        len--;
                                    }
                                    memcpy(*message, comment, len);
                                }
                            }
                        }
                    }
                    if(*comment && *message == NULL && grbl.on_gcode_comment)
                        *status = grbl.on_gcode_comment(comment);
                }
                comment = NULL;
                break

terjeio: The previous comment was somehow posted before I finished, trying again:

> Aha, maybe I should make a separate PR with just the string substitution-stuff, and move string registers to a plugin-repo?

I'll update the core, you add a plugin for string registers _and_ string substitution that includes string registers. This since I'll make comment handling completely overridable even for [MSG, ...), in gcode.c:

```
            case ')':
                if(comment && !gc_state.skip_blocks) {
                    *s1 = '\0';
                    if(!hal.driver_cap.no_gcode_message_handling) {

                        if(message && *message == NULL) {

                            if(grbl.on_process_gcode_comment)
                                *message = grbl.on_process_gcode_comment(comment);

                            if(*message == NULL) {
                                size_t len = s1 - comment - 3;

                                if(!strncasecmp(comment, "MSG,", 4) && (*message = malloc(len))) {

                                    comment += 4;
                                    while(*comment == ' ') {
                                        comment++;
                                        len--;
                                    }
                                    memcpy(*message, comment, len);
                                }
                            }
                        }
                    }
                    if(*comment && *message == NULL && grbl.on_gcode_comment)
                        *status = grbl.on_gcode_comment(comment);
                }
                comment = NULL;
                break;
```

`grbl.on_process_gcode_comment` will be defaulted to handle (PRINT, ..) and (DEBUG, ...) when expressions is enabled.
You can choose to do string replacements only and use the default handler for numeric ones, or replace the handler completely. Note that the `grbl.on_process_gcode_comment` signature does not allow for a status code return - this is deliberate since message comments should never fail with an error.

You may choose where to publish a string register plugin, you can add it to your own repo or make a PR for the [misc plugins repo](https://github.com/grblHAL/Plugins_misc/). The Web Builder can build with repos located outside grblHAL, but local builds cannot without manually adding your code to the source tree - at least not without adding it as a submodule which I am reluctant to do.

FYI I have already added string _parameter_ support to ngc_param.c (not yet comitted) in order to handle named o calls, you may build string registers on top of this or add it to your plugin. It is different from your string register code in that it does not return status codes and it will remove a string parameter on failure to allocate for it. Also, it will not reallocate on changes if the new length is less than the previous, this as an attempt to reduce heap fragmentation.

phansel:
@skasti have you seen https://www.klipper3d.org/Protocol.html?

PR #606, opened 2024-10-11
Add Setting for Axis Limitpos

jeefdumbo:
I have a suggestion for homing.
The negative home positions are not common for non-professional machines and are therefore rather confusing.
My machine, for example, has the Y motors and the limit switches on the 
Max Position.Setting the origin to 0 is therefore not applicable.
When reversing the homing direction I have the negative positions.
I therefore added a “Limit Pos” setting for the axes.
I can now define the position of the limit switches for each axis. This means it doesn't matter where the limit switches are mounted and I always have a defined position.
These changes may not be fully developed but will work.
Please excuse my English, I come from German-speaking countries.
I hope the pull request is correct. I very rarely work with git

### DIFF
--- a/config.h
+++ b/config.h
@@ -39,7 +39,7 @@ specific needs, i.e. performance tuning or adjusting to non-typical machines.
 If more than 3 axes are configured a compliant driver and board map file is needed.
  */
 #ifndef N_AXIS
-#define N_AXIS 3 // Number of axes
+#define N_AXIS 4 // Number of axes
 #endif
 
 /*! \def AXIS_REMAP_ABC2UVW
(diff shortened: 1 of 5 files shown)


skasti:
I also have my limit switches on the max-position, and the only "side-effect" of this is that my machines workspace is all in "negative space". And that is quite common as far as I understand 🤔

terjeio:
> Since it's not a big change

It is and I'll put this on my todo list. The main reason it is a big change is that you have added new data to the settings struct. This will cause a settings reset to defaults on an update - which I do not want to burden the users with too often. Also, I have a few other pending changes to the struct that I want to include when it is revised (and there is a revision number in the struct that has to be updated, along with all the drivers that checks it).

There is a few other issues with your PR as well, you have changed some default values in _config.h_ that should not be changed, and you have hijacked some existing setting numbers- new setting numbers has to be added instead.

PR #680, opened 2025-02-11
Small change to address spindle delay during tool change 

andrewmarles:
Not sure if this is the best way to fix this, but it looks like there is a disconcerting pause during semi-auto toolchanges when the spindle is off and the RPM is zero.


terjeio:
Not sure I get this. What is the actual state change? And is it for a PWM spindle where enable is controlled by RPM?

image

andrewmarles:
Seems with vfd spindles the value field is nonzero even when the spindle is off.  I had the same issue reading that field with the atc plug-in.

terjeio:
Oops, that is a bad one - there are other flags than on and ccw in the value. I'll add a function that checks only on and ccw.

andrewmarles:
This seems resolved with latest core updates.

PR #771, opened 2025-07-02
modbus-debug

andywwright:
Terje, thank you so much for this monumental work. I can't even begin to comprehend how many thousands of hours it takes. I really appreciate a huge chunk of your life being thrown on this project.

That being said, guys, I'm not a coder, but surely any debugging is better than none? To me trying to connect a new vfd with a 500 page manual without actually seeing what you're doing is like driving a car to an unfamiliar place blindfolded.

### DIFF
--- a/modbus_rtu.c
+++ b/modbus_rtu.c
@@ -21,10 +21,30 @@
 
 */
 
+
 #include <stdio.h>
 #include <string.h>
 
 #include "hal.h"
+#if MODBUS_DEBUG
+#include "stream.h"
+static void log_modbus_frame(const char *label,
+                             const uint8_t *data,
+                             uint8_t len)
+{
+    char buf[64];
+    size_t n = (size_t)snprintf(buf, sizeof(buf), "%s:", label);
+    hal.stream.write(buf);
+
+    for(uint_fast8_t i = 0; i < len; i++) {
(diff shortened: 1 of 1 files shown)


andywwright:
Thanks, changes made as suggested (I think)

andywwright:
btw the reason why I wanted MODBUS_DEBUG to be separate from DEBUG is because when iosender is connected to a modbus capable spindle, the console immediately gets flooded with the status messages like this

```
[MSG:Debug: TX: 01 03 00 C8 00 01 05 F4]
[MSG:Debug: RX: 01 03 02 00 00 B8 44]
```

at about 5Hz rate so not only any real DEBUG messages that you might hope for are immediately carried away, the simple console communication becomes a challenge.

terjeio:
I am closing this since the latest commit supports intercepting the Modbus stream handlers.
Utilizing that I have added a debug command in a [template plugin](https://github.com/grblHAL/Templates/tree/master/my_plugin/Modbus_command) that allows turning debugging on and off at run time. To further aid debugging Modbus commands can also be sent via a $ command implemented in the plugin or via the [G65P7 inbuilt macro](https://github.com/grblHAL/core/wiki/Expressions-and-flow-control#inbuilt-g65-macros).

PR #802, opened 2025-09-08
Add realtime report of distance-to-go 

engigeer:
I found myself missing a feature from industrial controls (e.g., FANUC) which show the remaining distance in the current travel move (i.e., currently executing block) as a distance-to-go readout. This PR adds support for computing this distance in the planner and a setting for adding it to the realtime report between the buffer and line numbers.

If you are open to it, I would also like to explore options for displaying these values in iosender, but that is likely a subject for a separate PR.

### DIFF
--- a/settings.h
+++ b/settings.h
@@ -631,6 +631,7 @@ typedef union {
         uint16_t machine_position   :1,
                  buffer_state       :1,
                  line_numbers       :1,
+                 distance_to_go     :1,
                  feed_speed         :1,
                  pin_state          :1,
                  work_coord_offset  :1,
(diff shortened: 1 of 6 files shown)


settings.h
```diff
@@ -631,6 +631,7 @@ typedef union {
         uint16_t machine_position   :1,
                  buffer_state       :1,
                  line_numbers       :1,
+                 distance_to_go     :1,
```
terjeio, line 634: Adding this in the middle of the value will force a settings revision and a settings reset on update, add it after `when_homing` instead to avoid that.

PR #858, opened 2025-11-25
Homing Alarm 11 Not Asserted Following E-Stop

engigeer:
I encountered an issue when homing required is enabled. It seems that Alarm 11 is never asserted following an e-stop with loss of position, but the machine remains in an Alarm state until the homing cycle is run. There is a message reported that homing is required, however these messages are not always visible (i.e., if I am operating the machine from a pendant / controller that only reports Alarm codes).

I'm thinking instead perhaps it would be better to mirror the handling case where homing is required but only some axes are homed.
https://github.com/grblHAL/core/blob/e272f773b558f9d9a8d37533c40ae45247fbf396/system.c#L466-L477

### DIFF
--- a/expanders_init.h
+++ b/expanders_init.h
@@ -31,7 +31,7 @@ extern void board_ports_init (void); // default is a weak function
 
 // I2C expanders
 
-#if PCA9654E_ENABLE || MCP3221_ENABLE || MCP4725_ENABLE
+#if PCA9654E_ENABLE || MCP3221_ENABLE || MCP4725_ENABLE || FLEXGPIO_ENABLE
 
 #if defined(I2C_ENABLE) && !I2C_ENABLE
 #undef I2C_ENABLE
@@ -54,6 +54,10 @@ extern void pca9654e_init(void);
 #endif
 
 // Third party I2C expander plugins goes after this line
+
+#if FLEXGPIO_ENABLE
+extern void flexgpio_init(void);
+#endif
 
 #endif // I2C expanders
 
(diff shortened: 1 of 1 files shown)


engigeer:
It appears that recent changes to the core have fixed this and I can now see the correct alarm state in ioSender. 

Ideally, any change in alarm code would also result in an update sent to a connected i2c display / keypad. Currently the state of the `machine_status_packet_t` struct is updated via `grbl.on_state_change`, and it seems this handler is not called if only the substate has changed?

I wonder if you would be open to modifying this behaviour, or perhaps this would be best addressed with a small change to the keypad plugin method for updating the struct?

terjeio:
I think the easiest option for now is to modify the keypad plugin since the substate is determined from data that is changed both inside and outside of the state machine, and even in interrupt context(s). If the current state is one that can have substates then just query the substate in `send_status_info()`.
To aid this a bit I will add add a new define:
`#define STATE_HAS_SUBSTATE (STATE_CYCLE|STATE_HOLD|STATE_ESTOP|STATE_ALARM|STATE_SAFETY_DOOR)`
that can be used to check which states that can have substates, e.g. something like this:
```
if(state & STATE_HAS_SUBSTATE) {
    substate = state_get_substate();
    ...
}
```

engigeer:
Apologies for the poor commit message. It seems I used my master branch to open this previous PR and then mistakenly pushed the changes here when testing the code you shared for PR #920.

In either case, the changes you shared worked perfectly and were ready for merge, so many thanks for your help. I will close #920 as complete.

PR #873, opened 2025-12-20
Fix Z home movement when XY already at TLS position

gcormier:
When X and Y coordinates are already at the tool length sensor (TLS) position during tool change restore, Z was incorrectly not moving to the home position after probing.

The issue was in the restore() function where Z movement was only executed inside a conditional block that checked if XY needed to move. This caused Z to remain at the probed position when starting from the TLS XY location.

Fixed by decoupling Z home movement from XY position comparison, ensuring Z always returns to home after probing (unless $346 disables full position restore).

If it is not desireable to always move to Z home, we will need to adjust this code.

### DIFF
--- a/tool_change.c
+++ b/tool_change.c
@@ -139,18 +139,20 @@ static void reset (void)
 // Restore coolant and spindle status, return controlled point to original position.
 static bool restore (void)
 {
-    bool ok;
+    bool ok = true;
     plan_line_data_t plan_data;
 
     plan_data_init(&plan_data);
     plan_data.condition.rapid_motion = On;
 
-    if(!(ok = (target.values[plane.axis_0] == previous.values[plane.axis_0] &&
-                target.values[plane.axis_1] == previous.values[plane.axis_1]))) {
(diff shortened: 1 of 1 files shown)


gcormier:
Compiled, flashed, working as intended.

https://photos.app.goo.gl/sfKYndjR3umDhWNy8

terjeio:
Good catch

karoria:
Did this bug run for tool change via tc.macro too? My Z coordinate comes inaccurate up to 0.5mm after tool change. My tool length sensor is accurate within 0.005mm. Even changed sensor with new one but still facing same issue. Can this be due to this bug? I make G43 activated always with $N0=G43 setting at machine startup.

terjeio:
> Did this bug run for tool change via tc.macro too?

No, the changed code is part of the "standard" tool change implementation and is completely switched out when _tc.macro_ is used.
Your  issue is either from another bug or from bad code in _tc.macro_. Perhaps related to the G43 at startup?

karoria:
Ok. Thanks for pointing. Will see more into it.

PR #887, opened 2026-01-09
Fix Z position tracking when stopping while parked

gcormier:
When executing a parking cycle ($41) and then stopping the job while
parked, the Z position was not being tracked correctly. This occurred
because sync_position() synchronized gc_state.position to the current
physical position (at parking height) rather than the logical work
position before parking.

Added state_await_parking_adjustment() function that adjusts gc_state
position back to the pre-park position when stopping while parked. The
adjustment is only applied when actual parking motion occurred
(park.flags.motion is set), so configurations where Z doesn't move
during parking are handled correctly.

There's probably many ways to tackle this, if you have other ways you would like to see it handled please let me know.

I have flashed and tested this on my machine and it fixes the behavior I was seeing.


gcormier:
@terjeio  Any thoughts on this?

terjeio:
I am not sure this is a good idea, the parser will get out of sync with the planner and the actual step position. I'll check it out when I am back home and have access to a machine.

gcormier:
I'm gonna close this - why don't I submit a new one that, upon hitting this edge case, the machine will reset the homed status of Z ? I can put it behind a firmware option if desireable?

gcormier:
Updated - $41 now has another bit to control this behavior (default to off, keep legacy functionality)

If enabled, when parked and stop, it will invalidate the parking axis homed status.

PR #914, opened 2026-02-22
PR for Modbus fixes around single coils and discrete inputs.

andrewmarles:
I think I may have found some small issues when working with single coils and the packed bytes that they use.  This seems to work against a nanoModbus server in my basic testing.

nanoModbus plugin here:
[https://github.com/Expatria-Technologies/grblhal_nmbs_plugin](https://github.com/Expatria-Technologies/grblhal_nmbs_plugin)

### DIFF
--- a/modbus.c
+++ b/modbus.c
@@ -221,6 +221,12 @@ FLASHMEM static void rx_packet (modbus_message_t *msg)
             case ModBus_Diagnostics:
                 response.num_values = 0; // TBA
                 break;
+            case ModBus_ReadCoils:
+            case ModBus_ReadDiscreteInputs:
+                response.num_values = min(msg->adu[2], 3); // byte count, not /2
+                for(idx = 0; idx < response.num_values; idx++)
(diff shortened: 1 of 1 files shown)


terjeio:
Thanks! Interesting plugin - grblHAL drivers has to get support for more serial ports?

andrewmarles:
Thank you, it has been my dream to do this for over a year now with PicoHAL and finally finding the time.  I am using it to drive an ATC.  I have pre-defined macros that map to things like "goto toolchange position" and "select pocket xxx."  With GRBLHAL we get all the motion stuff (homing, position, offsets etc) for free.

I have found it fairly seamless so far to work with the existing RS485 on FlexiHAL daisy-chained with the VFD.  I think we will rapidly fill up the bandwidth on Modbus though, but that is why the scripting is nice because you can send very minimal commands to nanoModbus that have very complex behaviour.

https://github.com/user-attachments/assets/c3b8fe50-3d0d-49f1-beaa-a9f76210194f

PR #920, opened 2026-03-08
Add core support for FLEXGPIO i2c Expander and Motor Alarm Expander Inputs

engigeer:
- Add init for **flexgpio** plugin to support use of i2c based RP2040 io expander from Expatria
- Add core support to facilitate defining and claiming per-axis motor fault pins by expander plugins
- Fix typo in aux input mask

### DIFF
--- a/ioports.c
+++ b/ioports.c
@@ -32,7 +32,9 @@
 #include "hal.h"
 #include "settings.h"
 
+#ifndef MAX_PORTS
 #define MAX_PORTS (Output_AuxMax - Output_Aux0 + 1)
+#endif
 
 typedef enum {
     Port_AnalogIn = 0,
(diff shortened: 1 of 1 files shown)


terjeio:
I have added an alternative solution that, IMO, is cleaner and easier to implement across drivers. I hope it is ok I did this without involving you...

Handling of the is still done in board specific code, main change driver side is that the pins to use is set via a call from the driver and that both expander and local pins can be used, even in a mix if required.
The [driver specific code for the Sienci SLB Ext ](https://github.com/grblHAL/STM32F4xx/blob/13f065c7017b6418c383586619b4699ce6e3dd7a/boards/longboard32.c#L151-L191)has been updated, this was the only board making use of separate inputs for each driver, making the decision to change the interface easier.

I have not added all your proposed changes, e.g. expanding the number of digital ports. The main reason for not doing this earlier is that adding port definitions eats into flash - and some drivers are a bit short on that.

Have you tested expander probe inputs? No latency issues?

engigeer:
> I have added an alternative solution that, IMO, is cleaner and easier to implement across drivers. I hope it is ok I did this without involving you...
> 
Sounds great. I will test with my board and confirm all is working as expected.
> 
> I have not added all your proposed changes, e.g. expanding the number of digital ports. The main reason for not doing this earlier is that adding port definitions eats into flash - and some drivers are a bit short on that.
> 
This makes sense, although maybe there could also be a way to expand the IO capability on boards/drivers that support it? I have not yet pinpointed the exact source of the issue I was fixing with this change, but I think it is related to having more pins than `AUX_MAX`, even though some of these pins are not `AUX` (e.g., limits). I’m wondering if maybe it is the loop in [map_reverse](https://github.com/grblHAL/core/blob/ee077d87305a125ed3a6840e995cd995f92e887a/ioports.c#L112)?

engigeer:
> Have you tested expander probe inputs? No latency issues?
>
Testing of this is still ongoing but I think there are some options if it proves to be too much. The expander was designed with a second irq pin connected to the primary mcu, which could be dedicated to probes. It seems like it should be possible to hook into the probe-select logic and then “hand-off” this signal between the various configured probes, depending on which is active.

PR #928, opened 2026-03-31
Fix -Werror=calloc-transposed-args with gcc 14.

bullestock:
gcc 14 introduces calloc-transposed-args, which checks that calls to calloc() conform to the "num, size" order of arguments.

Most of the calls in grblHAL are correct, but a few were not.

### DIFF
--- a/fs_device.c
+++ b/fs_device.c
@@ -43,7 +43,7 @@ bool device_fs_add_stream (io_stream_properties_t *stream)
     if(!stream->flags.claimable || !stream->flags.modbus_ready)
         return false;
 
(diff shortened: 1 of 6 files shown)


terjeio:
Hast the prototype changed as well? I get `extern void *calloc (size_t, size_t);` which is kind of ambiguous,

bullestock:
> Hast the prototype changed as well? I get `extern void *calloc (size_t, size_t);` which is kind of ambiguous,

According to cppreference.com, the prototype is `void* calloc( size_t num, size_t size )`.

PR #956, opened 2026-05-18
Add two new plugins

gcormier:
Two new plugins under development

1. grblhal-thermistor - https://github.com/gcormier/grblhal-thermistor
Read temperature of spindle via thermistor and analog input

2. grblhal-pneumaseal - https://github.com/gcormier/grblhal-pneumaseal
Intelligently manage air seals on spindle so we don't waste air all the time.

Not sure if you want to wait to merge these or not? I can develop locally first if that's better.

### DIFF
--- a/plugins_init.h
+++ b/plugins_init.h
@@ -253,6 +253,16 @@
   keyway_init();
 #endif
 
+#if PNEUMASEAL_ENABLE
+    extern void pneumaseal_init (void);
+    pneumaseal_init();
+#endif
+
+#if THERMISTOR_ENABLE
+    extern void thermistor_init (void);
+    thermistor_init();
+#endif
+
 // End third party plugin definitions.
 
 #if ODOMETER_ENABLE


terjeio:
I prefer to add links to your repos and make them available from the Web Builder as 3rd party plugins - so that I do not have to maintain them.

To make the plugins compatible with the RP2040 and ESP32 drivers you have to add a _CMakeLists.txt_ to each of them - similar to [this](https://github.com/grblHAL/Plugins_Bluetooth/blob/main/CMakeLists.txt).

gcormier:
Is there a web builder repo I can submit PR's to?

I will add cmake files.

terjeio:
>A Is there a web builder repo I can submit PR's to?

Not really - I have to update server scripts manually. The [main config file](https://github.com/grblHAL/drivers/blob/master/drivers.json) is public, but I will not accept PRs for that since changes to it may require server side updates.

To increase visibility you may add links to them in the [plugins readme](https://github.com/grblHAL/plugins#third-party-plugins).

gcormier:
CMake is added.

Do you want me to keep this PR open so the unique ID's are added to the plugins init?

PR #962, opened 2026-06-05
Spindle interlock plugin header

gcormier:


terjeio:
FYI I have moved you plugins to [Third party plugins](https://github.com/grblHAL/plugins#third-party-plugins) where they belong.

PR #966, opened 2026-06-16
Honor LITTLEFS_ENABLE=2 as YMODEM request when SDCARD is enabled

stevenrwood:
## Summary

Pre-existing code in `driver_opts.h` demotes `LITTLEFS_ENABLE` from `2` to `1` whenever `SDCARD_ENABLE` is also set (to avoid the root-mount conflict). The demotion runs **before** `FS_ENABLE` is computed, so the "`LITTLEFS_ENABLE=2` implies YMODEM upload" semantic — which works for LittleFS-alone — was unreachable in the combined SDCARD+LittleFS config. `[NEWOPT]` advertised `FS,SD` instead of `YM,SD`; `ymodem_init()` never fired.

## Fix

Capture the user's `=2` intent into a new `LITTLEFS_YMODEM_REQUESTED` flag **before** the demotion:

```c
#if LITTLEFS_ENABLE == 2 && SDCARD_ENABLE
#define LITTLEFS_YMODEM_REQUESTED   1
#undef LITTLEFS_ENABLE
#define LITTLEFS_ENABLE     1
#else
#define LITTLEFS_YMODEM_REQUESTED   0
#endif
```

The `FS_ENABLE` calculation in the combined case then grants YMODEM if `SDCARD_ENABLE==2` **or** `LITTLEFS_YMODEM_REQUESTED`. The mount-dir logic is tightened to put LittleFS at `/` only when SDCARD is absent at build time.

## Test plan

- [x] Build with `SDCARD_ENABLE=1, LITTLEFS_ENABLE=2, YMODEM_ENABLE`: `$I` reports `[NEWOPT:...,YM,SD,...]`, SOH/STX trigger fires `ymodem_init`, file lands on `/littlefs`.
- [x] Same config with `LITTLEFS_ENABLE=1`: `[NEWOPT:...,FS,SD,...]` (YM not advertised — unchanged).
- [x] LittleFS-alone (`LITTLEFS_ENABLE=2`, no SDCARD): YMODEM still fires (path unchanged).

---

Tested on grblHAL/iMXRT1062 (Teensy 4.1, BOARD_T41U5XBB).

### DIFF
--- a/driver_opts.h
+++ b/driver_opts.h
@@ -333,18 +333,16 @@
 #define SPINDLE3_ENABLE     0
 #elif SPINDLE3_ENABLE == -1 || SPINDLE3_ENABLE == SPINDLE_ALL || SPINDLE3_ENABLE == SPINDLE_ALL_VFD
 #warning "SPINDLE3_ENABLE cannot be set to -1, SPINDLE_ALL or SPINDLE_ALL_VFD"
-#undef SPINDLE3_ENABLE
-#define SPINDLE3_ENABLE     0
+#undef SPINDLE1_ENABLE
+#define SPINDLE1_ENABLE     0
 #endif
 
-#ifndef SPINDLE_ENABLE
 #if SPINDLE0_ENABLE == -1 || SPINDLE0_ENABLE == SPINDLE_ALL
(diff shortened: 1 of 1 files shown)


stevenrwood:
Withdrawing for now - reworking against my fork; may resubmit a refined version later. Thanks for the look.

PR #979, opened 2026-07-02
Fix feedrate calculation for rotary axes moves in G20 (imperial units)

engigeer:
It seems that the feed-rate used for movements involving only rotary axes is still scaled by a factor of 25.4 when imperial unit modal state (G20) is active.

The proposed correction here is only applied within the `ROTARY_FIX` code guard, but could make sense to apply in general. Also, since the change is applied at the planner level, the *corrected* feed-rate is not displayed in the parser state of ioSender, but I could not think of a way to do so which would not create potential issues for subsequent moves that did not provide a new F-code.

### DIFF
--- a/planner.c
+++ b/planner.c
@@ -497,24 +497,31 @@ bool plan_buffer_line (float *target, plan_line_data_t *pl_data)
 
     if(!block->condition.inverse_time &&
         !block->condition.rapid_motion &&
-         (motion.mask & settings.steppers.is_rotary.mask) &&
-          (motion.mask & ~settings.steppers.is_rotary.mask)) {
+        (motion.mask & settings.steppers.is_rotary.mask)){
+            
+        if(motion.mask & ~settings.steppers.is_rotary.mask) {
 
-        float linear_magnitude = 0.0f;
+            float linear_magnitude = 0.0f;
 
-        idx = 0;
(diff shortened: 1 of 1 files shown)


terjeio:
I am not sure about what to do with this, IMO needs a bit of thinking.
And I am quite busy with other stuff so not much free time for grblHAL.

engigeer:
> I am not sure about what to do with this, IMO needs a bit of thinking. And I am quite busy with other stuff so not much free time for grblHAL.

All good, completely understand. For now, I will keep testing this potential fix with my machine and see if I encounter any unexpected issues.

engigeer:
Assuming that a long-term fix might take some time, and since the current handling of rotary axes when `G20` is active (scaling `F` codes by a factor of **25.4** and applying the result as deg/min), is almost certainly not desirable.

Are there any changes I could make to this PR (e.g., additional code guards) that would make it a suitable candidate for an interim fix?

terjeio:
I have moved the ROTARY_FIX option to $701 and added an additional option to revert the feedrate to that in the [latest commit](https://github.com/grblHAL/core/commit/6523e08de8698123f808301b40de438c1ba5ce17) so closing this.
I am not sure a better solution exist, short of converting the core to imperial only?

PR #988, opened 2026-07-14
grbllib: flush the output stream buffer on reset, not just the input buffer

stevenrwood:
## Summary
`hal.stream.reset_read_buffer()` is called on every reboot to clear pending input, but there was no equivalent call for the output side. `reset_write_buffer` is defined in the HAL stream interface (optional, documented for Modbus/RS-485 support) and at least one driver (`Plugin_networking`'s `telnetd.c`) already implements it - but nothing in the core reboot sequence ever called it, on any transport.

**Consequence:** a stale, partially-transmitted TX message left in a driver's output buffer across a reset can leak out merged with the reboot's own output, with no separator - observed as a garbled/concatenated ALARM message (e.g. `Alarm:-1` instead of a clean `Alarm:3 - Reset/E-stop while in motion...`) immediately following a Reset-during-motion repro.

## Repro / verification
Reproduced identically on both Serial and Telnet on a Teensy 4.1 (iMXRT1062 driver) during a Reset-during-motion repro. After the fix, the `ALARM:` line prints cleanly on every repro attempt.

## Fix
Call `hal.stream.reset_write_buffer()` (guarded, since it's optional per the HAL interface) right alongside the existing `reset_read_buffer()` call in the reboot sequence.

## Test plan
- [x] Builds clean (Teensy 4.1 / PlatformIO)
- [x] Hardware-reproduced the corruption before the fix, and hardware-verified resolved after (identical on Serial and Telnet)

### DIFF
--- a/grbllib.c
+++ b/grbllib.c
@@ -484,6 +484,8 @@ FLASHMEM int grbl_enter (void)
 
         // Reset primary systems.
         hal.stream.reset_read_buffer();                 // Clear input stream buffer
+        if(hal.stream.reset_write_buffer)
+            hal.stream.reset_write_buffer();            // Clear output stream buffer
         gc_init(settings.flags.keep_offsets_on_reset);  // Set g-code parser to default state
         hal.limits.enable(settings.limits.flags.hard_enabled, (axes_signals_t){0});
         plan_reset();                                   // Clear block buffer and planner variables


terjeio:
The output buffer is not flushed to avoid garbled output so I need to be able to reproduce this myself.
BTW do you use ioSender and if so which version? Asking since it might be an issue with the sender.

stevenrwood:
I am using my own veraion of ioAwnswe (C2.01) which has diverged quite a bit from your version.  I you sre interested in the details, this file lays out all the changes.

stevenrwood:
Might help to include the link:  ioSender/Overview.pdf at integration * stevenrwood/ioSender<https://github.com/stevenrwood/ioSender/blob/integration/Overview.pdf>

It represents over 50,000 lines of new code written over the last 6 weeks.   Interestingly,  I didn't write or read a single line of code.  Claude Code did all the work.    I just described how I wanted to look and function.

I am a CNC notice, so  lot of my changes wee to make it easier for a novice to get up to speed.   Still have to finish the user manual with embedded videos showing how to use various features.   If you want to take a look, feedback welcome.

Steve

From: Steve Wood
Sent: Saturday, July 18, 2026 7:10 PM
To: grblHAL/core ***@***.***>
Subject: RE: [grblHAL/core] grbllib: flush the output stream buffer on reset, not just the input buffer (PR #988)

I am using my own veraion of ioAwnswe (C2.01) which has diverged quite a bit from your version.  I you sre interested in the details, this file lays out all the changes.

PR #989, opened 2026-07-18
Fix stack buffer overflows in ngc_flowctrl.c

94xhn:
This patch fixes two stack buffer overflow vulnerabilities inside \
gc_flowctrl.c\ where long user G-code subprogram/macro names retrieved via \
gc_string_param_get\ could overflow stack buffers \msg[100]\ and \ilename[60]\. 1. In \onNamedSubError\, the error message was formatted using \sprintf(msg, ...)\ where \msg\ is only 100 bytes long, while string parameters can be up to 255 bytes. 2. In the \NGCFlowCtrl_Call\ handler, macro filenames were formatted using \sprintf(filename, ...)\ where \ilename\ is only 60 bytes long. These are resolved by replacing unsafe \sprintf\ calls with bounds-checked \snprintf\.

### DIFF
--- a/ngc_flowctrl.c
+++ b/ngc_flowctrl.c
@@ -384,7 +384,7 @@ FLASHMEM static status_code_t onNamedSubError (status_code_t status)
                 char *name, msg[100];
                 closing = true;
                 if((name = ngc_string_param_get((ngc_string_id_t)o_label))) {
-                    sprintf(msg, "error %d in named sub %s.macro", (uint8_t)status, name);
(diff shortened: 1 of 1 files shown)


terjeio:
Instead of failing long names I have [instead changed the code](https://github.com/grblHAL/core/commit/c84cda043000a50a7c5f242f476f7e47487fde27#diff-eb24b309627070b15b4ae54eb0f446a4bb1212140a86720a7966bc31f48c8cad) to use heap to store them temporarily. Can you please confirm that this fixes your issue in a more graceful way? BTW be aware that the underlying file system may impose its own restrictions on filename lengths...

PR #994, opened 2026-07-23
Fix for reporting of updated overrides on reset

engigeer:
Currently it seems that if `$676` is configured to clear override values on reset, this change is not reported to a keypad display.

The proposed fix is just a single line addition in the reset loop to set the override reporting flag. This has been tested to work with the RP2040 driver and the latest version of the keypad plugin with I2C display (**DISPLAY_ENABLE=9**).

### DIFF
--- a/grbllib.c
+++ b/grbllib.c
@@ -481,6 +481,7 @@ FLASHMEM int grbl_enter (void)
             sys.override.control.parking_disable = settings.parking.flags.deactivate_upon_init;
 
         flush_override_buffers();
+        report_add_realtime(Report_Overrides);
 
         // Reset primary systems.
         hal.stream.reset_read_buffer();                 // Clear input stream buffer


grbllib.c
```diff
@@ -481,6 +481,7 @@ FLASHMEM int grbl_enter (void)
             sys.override.control.parking_disable = settings.parking.flags.deactivate_upon_init;
 
         flush_override_buffers();
+        report_add_realtime(Report_Overrides);
```
terjeio, line 484: Output overrides on changes only?

```
        if(!sys.cold_start && (sys.override.feed_rate != override.feed_rate || sys.override.rapid_rate != override.rapid_rate))
            report_add_realtime(Report_Overrides);

        do {
            if(spindle_is_enabled(--spindle_num)) {
                spindle_ptrs_t spindle = spindle_get(spindle_num);
                if(spindle->param->override_pct != DEFAULT_SPINDLE_RPM_OVERRIDE) {
                    spindle->param->override_pct = DEFAULT_SPINDLE_RPM_OVERRIDE;
                    report_add_realtime(Report_Overrides);
                }
            }
        } while(spindle_num);
```

engigeer: I think this is OK and would still address the issue I was observing, but I wonder what would be the need to limit so it only sends on changes?

It also seems that sending the message on every reset could maybe be a bit more robust to any other sort of communication errors. That way if there is an issue between the pendant display and ioSender, then issuing a reset command would re-sync things?

If you still prefer to limit it to changes only, I can of course update the PR according to your suggestion above.

terjeio: > I think this is OK and would still address the issue I was observing, but I wonder what would be the need to limit so it only sends on changes?

Not really needed but it is a change from before.

> It also seems that sending the message on every reset could maybe be a bit more robust to any other sort of communication errors. That way if there is an issue between the pendant display and ioSender, then issuing a reset command would re-sync things?

Then perhaps `Report_All` should be used instead? But that may break senders.

PR #995, opened 2026-07-24
Add read-only planner export API for external motion drivers.

ajayre:
Some drivers need access to the planned motion queue above the HAL stepper interface, for example to feed an external motion controller that does not share the same step/dir execution model. This adds a small, additive API that lets a driver or plugin read the current planner buffer without exposing the internal plan_block_t layout.

Changes:
- planner.h: add plan_export_block_t and plan_export_cursor_t, plus plan_get_revision(), plan_export_first() and plan_export_next().
- planner.c: add a volatile plan revision counter and plan_signal_changed() choke point, called whenever the buffer contents or computed velocity profile change (reset, queue, discard, re-plan). Export returns target, nominal velocity, exit velocity, acceleration, condition, line number, and a stable flag derived from block_buffer.planned.
- core_handlers.h: add grbl.on_planner_changed so subscribers can react when the plan changes. Keep the handler short; it may run from stepper prep context.

The export API is read-only and does not alter existing planner behaviour. No existing functions or structs were changed.

### DIFF
--- a/core_handlers.h
+++ b/core_handlers.h
@@ -88,6 +88,7 @@ typedef bool (*home_machine_ptr)(axes_signals_t cycle, axes_signals_t auto_squar
 typedef void (*on_parser_init_ptr)(parser_state_t *gc_state);
 typedef void (*on_state_change_ptr)(sys_state_t state);
 typedef void (*on_override_changed_ptr)(override_changed_t override);
+typedef void (*on_planner_changed_ptr)(void);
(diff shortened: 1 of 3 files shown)


ajayre:
This will allow me to write a driver for a CAN-bus based motion control system.

terjeio:
Would it not be better to replace st_prep_buffer() etc. with new code? Or do you intend to run the current step generation code in parallel?

BTW it is possible to hook into hal.stepper.pulse_start() and send per segment data to external drivers, IMO the data passed contains enough information to do that.

PR #1003, opened 2026-08-16
Core and shim changes to allow cleaner integration, only lightly tested

terjeio:
Two weak functions added which are overridden by the shim to hook into the core. G41/G42 support permanently added to the parser etc. based on your code, should allow your cutter compensation code to be added as a third party plugin and made available in the Web Builder if it works out.

What do you think?


terjeio:
Wrong target.

PR #1004, opened 2026-08-17
Fix for recursive call in onSettingsChanged.

hangsman:
### Summary
In `gcode.c`, `gc_init` chains the `on_settings_changed` callback by storing the previous handler into `settings_changed` and assigning `onSettingsChanged` to `grbl.on_settings_changed`. If `grbl.on_settings_changed` already references `onSettingsChanged` prior to chaining, `settings_changed` may end up pointing to `onSettingsChanged` itself, resulting in infinite recursion and a stack overflow when settings are updated.

### Changes
Added a check `settings_changed != onSettingsChanged` before invoking `settings_changed(settings, changed)` to prevent the callback from recursively calling itself.


terjeio:
When/how did the recursion happen?

hangsman:
The recursion is caused by a self-referencing hook created during soft-reset:

1. On cold boot (`grbl_enter` -> `gc_init`), `grbl.on_settings_changed` is still `NULL`, so `settings_changed` is assigned `NULL` while `grbl.on_settings_changed` becomes `onSettingsChanged` ([gcode.c:803-806](https://github.com/grblHAL/core/blob/master/gcode.c#L803-L806)).
2. When a sender connects and sends `0x18` (soft reset), `protocol_exec_rt_system` calls `gc_init(true)` again ([protocol.c:586](https://github.com/grblHAL/core/blob/master/protocol.c#L586)).
3. Since `settings_changed` is still `NULL`, `if(settings_changed == NULL)` passes again and executes `settings_changed = grbl.on_settings_changed`. But `grbl.on_settings_changed` is already `onSettingsChanged`, so `settings_changed` now points to itself.

Then any setting change calling `settings_changed(settings, changed)` enters an infinite recursion in `onSettingsChanged` ([gcode.c:755](https://github.com/grblHAL/core/blob/master/gcode.c#L755)).

terjeio:
Thanks, I see where now - all the boards I am testing with have I/O ports enabled and this sets `grbl.on_settings_changed` during boot masking this bug.

PR #1005, opened 2026-08-21
Fix out-of-bounds access in report_grbl_settings and add safety checks in settings.c

wyxun:
### Summary of Changes
1. **`report.c` (`report_grbl_settings`)**:
   - Fixes an out-of-bounds array access in the fallback iteration loop when `calloc` fails. The loop previously used the global total setting count (`n_settings`) instead of the group local count (`details->n_settings`), leading to memory corruption.
2. **`settings.c` (`setting_get_value` & `settings_iterator`)**:
   - Adds NULL and Thumb state sanity checks (`((uint32_t)setting->get_value & 1) == 0`) in `setting_get_value()` to prevent HardFault / INVSTATE crashes on ARM Cortex-M when iterating uninitialized or misaligned plugin setting callbacks during `$$` query.
   - Adds NULL check for `set` and `set->iterator` in `settings_iterator()`.

### DIFF
--- a/report.c
+++ b/report.c
@@ -507,7 +507,7 @@ FLASHMEM void report_grbl_settings (bool all, void *data)
         free(all_settings);
 
     } else do {
-        for(idx = 0; idx < n_settings; idx++)
+        for(idx = 0; idx < details->n_settings; idx++)
             settings_iterator(&details->settings[idx], print_setting, data);
     } while((details = details->next));
 }
--- a/settings.c
+++ b/settings.c
@@ -1923,6 +1923,11 @@ FLASHMEM char *setting_get_value (const setting_detail_t *setting, uint_fast16_t
 
             setting_id_t id = (setting_id_t)(setting->id + offset);
 
+            if(setting->get_value == NULL || ((uint32_t)(uintptr_t)setting->get_value & 1) == 0) {
+                value = (char *)"0";
+                break;
+            }
+
             switch(setting->datatype) {
 
                 case Format_Decimal:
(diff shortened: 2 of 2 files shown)


terjeio:
1. Good catch.
2. Are there any plugin code in the wild that has bad settings data?
IMO the `setting_get_value()` function should return `NULL` if the `get_value` pointer is missing for function based value access. Or what could (should?) be done instead is failing the registration if the data is bad...

wyxun:
I didn't encounter this with the official in-tree plugins, but rather during the development and testing of a custom user plugin (ATC / tool changing extensions). 

When registering custom function-based settings (e.g. `Setting_NonCoreFn`), if the `get_value` callback is accidentally omitted (`NULL`) during iterative bring-up, querying `$$` results in an indirect call to `0x0`, causing an immediate **HardFault / INVSTATE** crash on ARM Cortex-M.

Regarding the two approaches, **I recommend having `setting_get_value()` return `NULL`**:
- It is lightweight (a simple `if (!setting->get_value) return NULL;` check) and localized.
- It degrades gracefully: callers like `report_setting()` in `report.c` already check `if (value)` before outputting, preventing a HardFault without breaking the registration of other valid settings in the same group.
- Failing registration on an embedded target with no stdout console can make plugin development harder to debug.

I can update this PR with this change (and remove the Thumb-specific bit check) if that works for you

terjeio:
> I can update this PR with this change (and remove the Thumb-specific bit check) if that works for you

Please do so.

FYI I have added [some sanity checks](https://github.com/grblHAL/core/blob/351f77a80d9e625f981b9bcddea790163eee3026/settings.c#L2722-L2736) at registration in the latest commit.

wyxun:
I have updated this PR accordingly:
- `setting_get_value()` now simply breaks and returns `NULL` when `setting->get_value == NULL` (removed the MCU-specific check).
- Kept the `report_grbl_settings()` fallback iteration loop fix (`details->n_settings`).

